PR #1, opened 2023-08-17
ci: adapt workflows

### LINKED ISSUE #11: Update workflows

guidanoli:
Closes #11, #15, #16 and #18.


.github/.scopes.txt
ZzzzHui, line 1: do we continue using conventional commit, or we do changelog per PR?

guidanoli: Good question. I do think conventional commits forces us to organize commits better, even though they aren't used for automatic changelog generation. :-)

ZzzzHui: I thought we only use them for the purpose of auto generating changeling. 
I think it would be good to remove the boilerplate like `feat(contracts): ` if they are not used.

ZzzzHui (approved):

PR #26, opened 2023-08-26
feat: authority factory

### LINKED ISSUE #25: Add `Authority` factory
## 📚 Context

Every DApp must have a consensus in order to validate outputs against claims. Currently, we only have implemented the Authority consensus model. In order for the DApp developer to deploy an Authority contract, they must possess the source code, or at least the creation code of the contract (which can be found in some JSONs output by the Solidity compiler). This way of deploying an Authority contract may, however, be inconvenient for some projects that do not have Hardhat or framework alike to handle contract deployment. This issue is clearly crucial.

## ✔️ Solution

Create a contract that deploys `Authority` contracts, ergo an `Authority` factory.
There would also have to be a script that deploys such factory.

guidanoli:
Closes #25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `abi` directory to the `@cartesi/rollups` package, ideal for language bindings.
+- `AuthorityFactory`: Allows anyone to deploy `Authority` contracts. Supports deterministic deployment.
 
 ### Removed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `AuthorityFactory`: Allows anyone to deploy `Authority` contracts. Supports deterministic deployment.
+
```
xdaniortega, line 13: Should we add this comments in the `Changelog` manually again?

xdaniortega (approved):

CHANGELOG.md
guidanoli: Yes, we've opted to change the `CHANGELOG` manually per PR.

@gligneul has realized that we can't rely on conventional commits only. There is still manual work of filtering and tailoring commit messages. So, it's better to spread this work per PR rather than concentrate it all in release time.

pedroargento (approved):

PR #28, opened 2023-08-26
feat: history factory

### LINKED ISSUE #27: Add `History` factory
## 📚 Context

Currently, if one wants to deploy a `History` contract, the easiest way is through smart contract frameworks like Hardhat or Foundry. Still, this is not trivial for most web3 users, thus limiting the number of validators that can deploy their own Consensus contracts and associated History contracts. Furthermore, the DApp developer has to trust that the validator didn't deploy a malicious implementation of a consensus contract.

## ✔️ Solution

We assume that validators are able to sign and submit transactions to the blockchain. Therefore, a safe and easy way for them to deploy a History contract is by submitting a message call to a factory contract that deploys a new History contract.

## ⛓ Requires

- [x] cartesi/rollups#68

## 📈 Subtasks

- [x] Define interface
- [x] Implement interface
- [x] Add test cases for contract
- [x] Create deployment script for contract

guidanoli:
Closes #27

### DIFF
--- a/onchain/rollups/contracts/history/HistoryFactory.sol
+++ b/onchain/rollups/contracts/history/HistoryFactory.sol
@@ -1,0 +1,50 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+pragma solidity ^0.8.8;
+
+import {Create2} from "@openzeppelin/contracts/utils/Create2.sol";
+
+import {IHistoryFactory} from "./IHistoryFactory.sol";
+import {History} from "./History.sol";
+
+/// @title History Factory
+/// @notice Allows anyone to reliably deploy a new `History` contract.
+contract HistoryFactory is IHistoryFactory {
+    function newHistory(
+        address _historyOwner
+    ) external override returns (History) {
+        History history = new History(_historyOwner);
+
+        emit HistoryCreated(_historyOwner, history);
+
+        return history;
+    }
+
+    function newHistory(
+        address _historyOwner,
+        bytes32 _salt
+    ) external override returns (History) {
+        History history = new History{salt: _salt}(_historyOwner);
+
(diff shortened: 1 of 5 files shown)


ZzzzHui (dismissed):

onchain/rollups/test/foundry/history/HistoryFactory.t.sol
```diff
@@ -0,0 +1,96 @@
...
+        // Precalculated address must match actual address
+        assertEq(precalculatedAddress, address(history));
+
+        precalculatedAddress = factory.calculateHistoryAddress(
+            _historyOwner,
+            _salt
+        );
+
+        // Precalculated address must STILL match actual address
+        assertEq(precalculatedAddress, address(history));
```
xdaniortega, line 90: aren't we comparing here the same as in lines 70-73 and 82? Which is the case it can be different?

onchain/rollups/contracts/history/HistoryFactory.sol
```diff
@@ -0,0 +1,50 @@
...
+        bytes32 _salt
+    ) external override returns (History) {
+        History history = new History{salt: _salt}(_historyOwner);
+
+        emit HistoryCreated(_historyOwner, history);
+
+        return history;
+    }
+
+    function calculateHistoryAddress(
```
xdaniortega, line 35: Why do we need this method? Why is it better to perform this calculation onchain than offchain?

guidanoli: This calculation can happen entirely off-chain. Remember, this is a `view` function. No transaction needs to be sent to the blockchain. This method has several benefits for DApp developers. You can know the address of the contract before even deploying it. :-)

xdaniortega (approved):

pedroargento (dismissed):

pedroargento (approved):

PR #43, opened 2023-08-28
feat!: output unification

### LINKED ISSUE #42: Output Unification
## 📚 Context

Currently, the off-chain machine is able to generate two types of outputs: vouchers and notices. On-chain, these outputs can be validated by anyone with the help of validity proofs.

The proof consists on reconstructing the epoch hash and comparing it with the epoch hash that was claimed by the DApp's consensus. If they are equal, then we can be sure the output is valid.

Currently, this epoch hash is composed by the following components.

* The machine Merkle root hash
* The voucher Merkle root hash
* The notice Merkle root hash

As you can see, vouchers and notices each have their own separate Merkle tree. This helps to avoid misinterpreting notices as vouchers and vice versa.

There are, however, two main issues with this design. First, it unnecessarily complicates the on-chain and off-chain code. Second, it hinders the implementation of new output types, since we would have to add a new Merkle tree, which would result in a breaking change in the validation procedure.

## ✔️ Solution

And so, we noticed that things would be much simpler if we had a single tree for all verifiable outputs. To differentiate between output types, we could add a prefix to each output. For example, `0x00` for vouchers and `0x01` for notices. As a result, the epoch hash would only be composed of two components:

* The machine Merkle root hash
* The output Merkle root hash

guidanoli:
Closes #42


onchain/rollups/contracts/library/LibOutputValidation.sol
```diff
@@ -105,7 +190,7 @@ library LibOutputValidation {
         // is contained in it. We can't simply use hashOfOutput because the
         // log2size of the leaf is three (8 bytes) not  five (32 bytes)
         bytes32 merkleRootOfHashOfOutput = MerkleV2.getMerkleRootFromBytes(
-            abi.encodePacked(keccak256(encodedOutput)),
+            abi.encodePacked(keccak256(abi.encode(output))),
```
guidanoli, line 192: **Note:** this is a temporary solution to generate correct output validity proofs. We would like to hash `output` directly, without having to ABI-encode it.

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
```diff
@@ -702,15 +702,28 @@ contract CartesiDAppTest is TestBase {
...
+        return OutputEncoding.encodeNotice(notice);
+    }
+
     function writeInputs() internal {
         for (uint256 i; i < outputEnums.length; ++i) {
             LibServerManager.OutputEnum outputEnum = outputEnums[i];
             if (outputEnum == LibServerManager.OutputEnum.VOUCHER) {
                 Voucher memory voucher = getVoucher(i);
-                writeInput(i, voucher.destination, voucher.payload);
+                writeInput(i, noticeSender, this.encodeVoucher(voucher));
```
guidanoli, line 723: **Note:** This is a temporary solution to generate correct proofs. The new echo DApp should allow us to generate whatever vouchers and notices we need, as well as any output we might come up with in the future.

onchain/rollups/contracts/library/LibOutputValidation.sol
```diff
@@ -5,10 +5,102 @@ pragma solidity ^0.8.8;
...
+///          xxx         │
+///         xxxxx        │
+///        xxxxxxx       │
+///       xxxxxxxxx      ├──> Epoch's outputs Merkle tree (H)
+///      xxxxxxxxxxx     │
+///     xxxxxxxxxxxxx    │
+///    xxxxxxxxxxxxxxx   │
+///   xxxxxxxxxxxxxxxxx
+/// ┌────────┬─┬────────┐
+/// │   ...  │┼│  ...   │ ───> For each input in the epoch (G)
```
ZzzzHui, line 66: Before rebase, `(G)` appears again in the struct `OutputValidityProof`'s comments, i.e.
```
/// @param inputIndexWithinEpoch Which input, inside the epoch, the output belongs to (G)
```
But because the struct `OutputValidityProof` is moved to a new file after rebase, this reference is also removed

onchain/rollups/test/foundry/dapp/Application.t.sol
```diff
@@ -978,8 +992,7 @@ contract ApplicationTest is TestBase {
                 inputIndexWithinEpoch: v.inputIndexWithinEpoch.toUint64(),
                 outputIndexWithinInput: v.outputIndexWithinInput.toUint64(),
                 outputHashesRootHash: v.outputHashesRootHash,
-                vouchersEpochRootHash: v.vouchersEpochRootHash,
-                noticesEpochRootHash: v.noticesEpochRootHash,
+                outputsEpochRootHash: v.noticesEpochRootHash,
```
ZzzzHui, line 995: Once the off-chain component is ready, we should update this and encoding and the helper library `LibServerManager`

ZzzzHui (approved):

onchain/rollups/test/foundry/dapp/Application.t.sol
guidanoli: Yes, many lines in this PR are mere artifacts that let us test the new output validation scheme using the old machine.

PR #45, opened 2023-08-28
feat!: allow vouchers to transfer Ether

### LINKED ISSUE #44: Payable Vouchers
## 📚 Context

Currently, vouchers allow DApps to interact with contracts in the base layer through arbitrary message calls. Each voucher is composed of two fields: the destination address and the payload. If the destination address contains a Solidity contract, then the payload encodes a function call (the function selector + the ABI-encoded arguments).

$$
Voucher = Destination \times Payload
$$

However, smart contracts have the ability to pass some amount of Wei along with the message call to the destination address. See the `value` argument of the `CALL` opcode[^2]. Currently, vouchers are passing zero Wei along with the message call, but there are several examples of functions that may require some Wei to be sent along. Take the `deposit()` function in the Wrapped ETH token contract[^3], for example. When someone calls this function, the smart contract mints as many tokens as the amount of Wei passed along the call, which can be accessed by the `msg.value` keyword. This guarantees that WETH tokens is backed by real Ether.

This is just one example, but there are several others. For example, contracts may require some Wei to be sent along as a form of internal fee. Another interesting use case is for Ether withdrawals. If a DApp wishes to withdraw some amount of Ether, they could do so through a message call destined to the Ether recipient, with an empty payload, and the value in Wei to be withdrawn. The recipient can be either an EOA[^4] or a smart contract, in which case the transfer may trigger a fallback function. With payable vouchers, there would be no need for the `withdrawEther` function[^5]. In fact, Ether withdrawals through payable vouchers would certainly consume less gas.

### Updating vouchers

We can add a new field to vouchers: the amount of Wei[^1] to be sent along the message call.

$$
Voucher' = Destination \times Payload \times Value
$$

To keep backwards compatibility with the previous voucher structure, we may keep old entry points while using the new voucher structure with the default value of zero Wei. This appears to be the easiest option to implement, given that we don't need to create any new output type.

### Adding a new output type

We may also choose to add a new output type, with the format described above, keeping full backwards compatibility with the current voucher structure. This would result in a larger `CartesiDApp` contract, and thus, make the DApp deployment cost higher.

### Implementing payable vouchers as a library

If we implement a type of output that does an arbitrary `DELEGATECALL`, we could simply implement a library that executes a payable voucher. The DApp would then need to emit a `DELEGATECALL` voucher destined to this library, with the voucher fields encoded as function arguments.

[^1]: the smallest denomination of Ether
[^2]: https://ethervm.io/#F1
[^3]: https://github.com/WETH10/WETH10/blob/87ec30256dab62459a2e6d2a1741b44d345881f1/contracts/WETH10.sol#L86-L90
[^4]: Externally-owned account
[^5]: https://github.com/cartesi/rollups/blob/495a12c5cfc3342b005dc82d9916f4b0536cff57/onchain/rollups/contracts/dapp/CartesiDApp.sol#L239-L250

guidanoli:
Closes #44 and #91

### DIFF
--- a/onchain/rollups/test/foundry/dapp/Application.t.sol
+++ b/onchain/rollups/test/foundry/dapp/Application.t.sol
@@ -53,6 +53,7 @@ contract ApplicationTest is TestBase {
 
     struct Voucher {
         address destination;
+        uint256 value;
         bytes payload;
     }
 
@@ -671,9 +672,17 @@ contract ApplicationTest is TestBase {
     }
 
     function _addVoucher(address destination, bytes memory payload) internal {
+        _addVoucher(destination, 0, payload);
+    }
+
+    function _addVoucher(
+        address destination,
+        uint256 value,
+        bytes memory payload
+    ) internal {
         uint256 index = _outputEnums.length;
         _outputEnums.push(LibServerManager.OutputEnum.VOUCHER);
-        _vouchers[index] = Voucher(destination, payload);
+        _vouchers[index] = Voucher(destination, value, payload);
     }
 
     function _addNotice(bytes memory notice) internal {
@@ -712,7 +721,7 @@ contract ApplicationTest is TestBase {
         return
             abi.encodeCall(
                 Outputs.Voucher,
-                (voucher.destination, voucher.payload)
+                (voucher.destination, voucher.value, voucher.payload)
             );
     }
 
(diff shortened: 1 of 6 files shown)


onchain/rollups/test/foundry/dapp/Application.t.sol
```diff
@@ -441,79 +442,6 @@ contract ApplicationTest is TestBase {
         _executeVoucher(voucher, proof);
     }
 
-    function testWithdrawEtherContract(
-        uint256 value,
-        address notApplication
-    ) public {
-        vm.assume(value <= address(this).balance);
-        vm.assume(notApplication != address(_app));
-        address receiver = address(new EtherReceiver());
```
ZzzzHui, line 450: Can remove the declaration of `EtherReceiver` contract in this test contract, since it's not used anywhere else

ZzzzHui (approved):

PR #47, opened 2023-08-28
feat!: encode inputs with strict mode

### LINKED ISSUE #46: Inputs added by input relays are hard to decode
## 📚 Context

Let us start with a short introduction on inputs. In order to interact with a DApp in a verifiable way, one must send an input to the `InputBox` contract. Off-chain, the node eventually picks up this new input by listening to `InputAdded` events emitted by this contract, and forwards it to the Cartesi Machine running the DApp back-end.

Whenever a DApp receives an input, it generally follows these steps: check if the input sender and/or the input format are sound; then apply the effects of accepting the input. In some cases, the DApp only checks whether the input comes from a trusted smart contract and assumes the input format is correct. In other cases, the DApp assumes the input comes from an EOA (externally-owned account) and only checks if the input is properly formatted. For example, it may check whether the input is a valid JSON string, and implements a given schema. 

Input relays are smart contracts that relay information pertaining to the base layer to applications running on the execution layer through inputs. Portals, for example, relay information about asset transfers. Given that input relays are meant to be used by all applications, **we should make an effort to make the inputs added by them simple to decode**.

Currently, the input relays encode inputs with Solidity's [non-standard packed mode](https://docs.soliditylang.org/en/v0.8.19/abi-spec.html#non-standard-packed-mode). Compared to the [strict encoding mode](https://docs.soliditylang.org/en/v0.8.19/abi-spec.html#strict-encoding-mode), which is ubiquitously used by Solidity, the packed mode yields smaller binary outputs, which is interesting from a gas consumption point-of-view. The problem with this mode is that, for some schemas, the encoding is not injective, meaning there is no associated decode function.

However, even though the input relays use schemas for which there are associated decode functions, this decoding algorithm is not generally defined by libraries that manipulate data encoded according to the Solidity ABI. Instead, they generally only define encode/decode entry points for the strict encoding mode. The DApp developer, therefore, has to implement the decoding algorithm themselves, which may contain vulnerabilities that can be exploited by bad actors.

## ✔️ Solution

To promote a smoother developer experience with code reusability, and to avoid ad-hoc, error-prone implementations of important algorithms, we would like to switch the encoding of inputs added by input relays from the non-standard packed mode to the strict encoding mode.

## :people_holding_hands: Acknowledgements

Thanks, @fmoura @miltonjonat @diegonehab @GCdePaula, for pointing out this problem to me.

guidanoli:
Closes #46

### DIFF
--- a/onchain/.changeset/silver-toys-sell.md
+++ b/onchain/.changeset/silver-toys-sell.md
@@ -1,0 +1,7 @@
+---
+"@cartesi/rollups": major
+---
+
+Updated encoding of inputs added by portals and relays to use `abi.encode` instead of `abi.encodePacked`.
+This change was made to simplify the decoding of inputs into its fields through the `abi.decode` function.
+DApp developers should update their input decoding code to use `abi.decode` for inputs added by portals and relays.
--- a/onchain/rollups/contracts/common/InputEncoding.sol
+++ b/onchain/rollups/contracts/common/InputEncoding.sol
@@ -22,12 +22,7 @@ library InputEncoding {
         uint256 value,
         bytes calldata execLayerData
     ) internal pure returns (bytes memory) {
-        return
-            abi.encodePacked(
-                sender, //              20B
-                value, //               32B
-                execLayerData //        arbitrary size
-            );
+        return abi.encode(sender, value, execLayerData);
     }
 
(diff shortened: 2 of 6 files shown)


guidanoli:
Not planned

PR #49, opened 2023-08-28
feat: add quorum

### LINKED ISSUE #48: Design a contract that implements a Quorum consensus model
## 📚 Context

Currently, the only consensus model we've implemented is [`Authority`](https://github.com/cartesi/rollups/blob/main/onchain/rollups/contracts/consensus/authority/Authority.sol).
One major issue with this consensus model, however, is that it is controlled by a single address.
This address can perform several tasks instantly: submit claims, change history, etc.

## ✔️ Solution

We'd like to implement a Quorum consensus model, which is controlled by $n \in \mathbb{N}$ validators, such that it is always possible to reach consensus, when performing any task, assuming at least one validator is active. Some details are still open for discussion:

- How would you deterministically set the epoch boundaries?
- According to your criterion, how would you deal with empty epochs[^1]?
- Would the validator set be dynamic? If so, how would it change?
- Would the Quorum be composable with other consensus models? If so, how?
- How would the Quorum distribute ERC-20 tokens amongst validators?
- Would the Quorum have discreet phases? If so, how would the state machine look like?

[^1]: i.e. epochs that have zero inputs

guidanoli:
Closes #48

### DIFF
--- a/onchain/rollups/contracts/consensus/quorum/Quorum.sol
+++ b/onchain/rollups/contracts/consensus/quorum/Quorum.sol
@@ -1,0 +1,153 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+pragma solidity ^0.8.8;
+
+import {BitMaps} from "@openzeppelin/contracts/utils/structs/BitMaps.sol";
+
+import {AbstractConsensus} from "../AbstractConsensus.sol";
+import {InputRange} from "../../common/InputRange.sol";
+
+/// @notice A consensus model controlled by a small, immutable set of `n` validators.
+/// @notice You can know the value of `n` by calling the `numOfValidators` function.
+/// @notice Upon construction, each validator is assigned a unique number between 1 and `n`.
+/// These numbers are used internally instead of addresses for optimization reasons.
+/// @notice You can list the validators in the quorum by calling the `validatorById`
+/// function for each ID from 1 to `n`.
+contract Quorum is AbstractConsensus {
+    using BitMaps for BitMaps.BitMap;
+
+    /// @notice The total number of validators.
+    /// @notice See the `numOfValidators` function.
+    uint256 private immutable _numOfValidators;
+
+    /// @notice Validator IDs indexed by address.
+    /// @notice See the `validatorId` function.
+    /// @dev Non-validators are assigned to ID zero.
+    /// @dev Validators have IDs greater than zero.
+    mapping(address => uint256) private _validatorId;
+
+    /// @notice Validator addresses indexed by ID.
+    /// @notice See the `validatorById` function.
+    /// @dev Invalid IDs map to address zero.
+    mapping(uint256 => address) private _validatorById;
+
+    /// @notice Votes in favor of a particular claim.
+    /// @param inFavorCount The number of validators in favor of the claim
+    /// @param inFavorById The set of validators in favor of the claim
+    /// @dev `inFavorById` is a bitmap indexed by validator IDs.
+    struct Votes {
+        uint256 inFavorCount;
+        BitMaps.BitMap inFavorById;
+    }
+
+    /// @notice Votes indexed by claim
+    /// (application address, first input index, last input index, and epoch hash).
+    /// @dev See the `numOfValidatorsInFavorOf` and `isValidatorInFavorOf` functions.
+    mapping(address => mapping(uint256 => mapping(uint256 => mapping(bytes32 => Votes))))
+        private _votes;
+
+    /// @param validators The array of validator addresses
+    /// @dev Duplicates in the `validators` array are ignored.
+    constructor(address[] memory validators) {
+        uint256 n;
+        for (uint256 i; i < validators.length; ++i) {
+            address validator = validators[i];
+            if (_validatorId[validator] == 0) {
+                uint256 id = ++n;
+                _validatorId[validator] = id;
+                _validatorById[id] = validator;
+            }
+        }
+        _numOfValidators = n;
+    }
+
+    /// @notice Submit a claim.
+    /// @notice If the majority of the quorum submit a claim, it is accepted.
+    /// @param application The application address
+    /// @param r The input range
+    /// @param epochHash The epoch hash
+    /// @dev Can only be called by a validator.
+    function submitClaim(
+        address application,
+        InputRange calldata r,
+        bytes32 epochHash
+    ) external {
+        uint256 id = _validatorId[msg.sender];
+        require(id > 0, "Quorum: caller is not validator");
+
+        emit ClaimSubmission(msg.sender, application, r, epochHash);
+
+        Votes storage votes = _getVotes(application, r, epochHash);
+
+        if (!votes.inFavorById.get(id)) {
+            votes.inFavorById.set(id);
+            if (++votes.inFavorCount == 1 + _numOfValidators / 2) {
+                _acceptClaim(application, r, epochHash);
+            }
+        }
+    }
+
+    /// @notice Get the number of validators.
+    function numOfValidators() external view returns (uint256) {
+        return _numOfValidators;
+    }
+
+    /// @notice Get the ID of a validator.
+    /// @param validator The validator address
+    /// @dev Validators have IDs greater than zero.
+    /// @dev Non-validators are assigned to ID zero.
+    function validatorId(address validator) external view returns (uint256) {
+        return _validatorId[validator];
+    }
+
+    /// @notice Get the address of a validator by its ID.
+    /// @param id The validator ID
+    /// @dev Validator IDs range from 1 to `N`, the total number of validators.
+    /// @dev Invalid IDs map to address zero.
+    function validatorById(uint256 id) external view returns (address) {
+        return _validatorById[id];
+    }
+
+    /// @notice Get the number of validators in favor of a claim.
+    /// @param application The application address
+    /// @param r The input range
+    /// @param epochHash The epoch hash
+    /// @return Number of validators in favor of claim
+    function numOfValidatorsInFavorOf(
+        address application,
+        InputRange calldata r,
+        bytes32 epochHash
+    ) external view returns (uint256) {
+        return _getVotes(application, r, epochHash).inFavorCount;
+    }
+
+    /// @notice Check whether a validator is in favor of a claim.
+    /// @param application The application address
+    /// @param r The input range
+    /// @param epochHash The epoch hash
+    /// @param id The ID of the validator
+    /// @return Whether validator is in favor of claim
+    /// @dev Assumes the provided ID is valid.
+    function isValidatorInFavorOf(
+        address application,
+        InputRange calldata r,
+        bytes32 epochHash,
+        uint256 id
+    ) external view returns (bool) {
+        return _getVotes(application, r, epochHash).inFavorById.get(id);
+    }
+
+    /// @notice Get a `Votes` structure from storage from a given claim.
+    /// @param application The application address
+    /// @param r The input range
+    /// @param epochHash The epoch hash
+    /// @return The `Votes` structure related to given claim
+    function _getVotes(
+        address application,
+        InputRange calldata r,
+        bytes32 epochHash
+    ) internal view returns (Votes storage) {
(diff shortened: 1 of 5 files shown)


guidanoli:
# Benchmark

For future optimizations, I've run `forge test` with some extra options.

```sh
forge test --match-contract Quorum --gas-report --fuzz-seed 1
```

Here is the gas report.

| Quorum |                 |       |        |       |         |
|-------------------------------------------------------|-----------------|-------|--------|-------|---------|
| Deployment Cost                                       | Deployment Size |       |        |       |         |
| 1763700                                               | 9019            |       |        |       |         |
| Function Name                                         | min             | avg   | median | max   | # calls |
| getHistory                                            | 280             | 280   | 280    | 280   | 1       |
| submitClaim                                           | 48629           | 56786 | 49200  | 72529 | 3       |

The most important data points are the **min** and **max** gas costs for the **submitClaim** function.

ZzzzHui:
This branch could use a rebase

ZzzzHui:
should the commit "feat: revert if erc20 deposit failed" be in this PR? Maybe you rebased this PR on "next" branch but pushed to base "main"?

guidanoli:
> should the commit "feat: revert if erc20 deposit failed" be in this PR? Maybe you rebased this PR on "next" branch but pushed to base "main"?

Oops, sorry! I think I rebased this on top of `next`. Let me fix that real quick... Thanks!

onchain/rollups/test/foundry/consensus/quorum/Quorum.t.sol
```diff
@@ -0,0 +1,270 @@
...
+        // resubmitting makes no difference
+        submitClaim(quorum, claim, submitLog, 1);
+        assertEq(history.numOfClaims(), 0);
+
+        submitClaim(quorum, claim, submitLog, 2);
+        assertEq(history.numOfClaims(), 1);
+        assertEq(history.claim(0), claim);
+
+        submitClaim(quorum, claim, submitLog, 3);
+        assertEq(history.numOfClaims(), 1);
```
ZzzzHui, line 205: we could make this into a `for` loop if we wanna test some quorums with larger size. Because BitMaps uses a new storage slot after, in our case, the first 255 bit positions being used. So quorum with size at least 256 would be ideal.

guidanoli: Good point. Let me add a test for 256 validators. :-)

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -0,0 +1,152 @@
...
+pragma solidity ^0.8.8;
+
+import {BitMaps} from "@openzeppelin/contracts/utils/structs/BitMaps.sol";
+
+import {AbstractConsensus} from "../AbstractConsensus.sol";
+import {InputRange} from "../../common/InputRange.sol";
+
+/// @notice A consensus model controlled by a small, immutable set of `n` validators.
+/// @notice You can know the value of `n` by calling the `numOfValidators` function.
+/// @notice Upon construction, each validator is assigned a unique number from 1 and `n`.
```
pedroargento, line 13: `from 1 to n` or `between 1 and n`

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -0,0 +1,152 @@
...
+        BitMaps.BitMap inFavorById;
+    }
+
+    /// @notice Votes indexed by claim.
+    /// @dev See the `numOfValidatorsInFavorOf` and `isValidatorInFavorOf` functions.
+    mapping(address => mapping(uint256 => mapping(uint256 => mapping(bytes32 => Votes))))
+        private _votes;
+
+    /// @param validators The array of validator addresses
+    /// @dev Duplicates in the `validators` array are ignored.
```
pedroargento, line 50: Is it worth to pay the gas for this logic to ignore repeated validators in the array? Maybe this can be handled by the script/frontend/individual that calls the QuorumFactory.

The consequences of repeated validators is a broken Quorum, but it shouldnt affect anyone unless a DApp dont know its broken.

onchain/rollups/contracts/consensus/quorum/Quorum.sol
guidanoli: Maybe `between 1 and n`, right?

onchain/rollups/contracts/consensus/quorum/Quorum.sol
guidanoli: Given that the quorum will be generally small, and that this code will only be executed once by whoever deploys the `Quorum` contract, I think it is worth.

guidanoli: After some benchmarking, it seems that we are only wasting 0.1% more gas, so I think it is still worth it.

onchain/rollups/contracts/consensus/quorum/Quorum.sol
pedroargento: I actualy like `from 1 to n` because its how they are ordered and assigned.` between 1 and n` doesn't convey order and its IMO.

onchain/rollups/test/foundry/consensus/quorum/Quorum.t.sol
```diff
@@ -0,0 +1,293 @@
...
+        Quorum quorum = _deployQuorum(numOfValidators);
+        bool[] memory submitted = new bool[](numOfValidators + 1);
+        for (uint256 id = 1; id <= numOfValidators; ++id) {
+            _submitClaimAs(quorum, claim, id);
+            submitted[id] = true;
+            _checkSubmitted(quorum, claim, submitted);
+        }
+    }
+
+    function testSubmitClaim300(Claim calldata claim) external {
```
ZzzzHui, line 177: just out of curiosity, what this 300 means :)

onchain/rollups/test/foundry/consensus/quorum/Quorum.t.sol
```diff
@@ -0,0 +1,293 @@
...
+
+        Quorum quorum = _deployQuorum(numOfValidators);
+
+        uint256 id = numOfValidators;
+
+        vm.prank(quorum.validatorById(id));
+        quorum.submitClaim(claim);
+
+        assertTrue(quorum.isValidatorInFavorOf(claim, id));
+        assertEq(quorum.numOfValidatorsInFavorOf(claim), 1);
```
ZzzzHui, line 188: we could also call `_submitClaimAs` instead, but the current lines work too

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -0,0 +1,152 @@
...
+
+    /// @notice Validator IDs indexed by address.
+    /// @notice See the `validatorId` function.
+    /// @dev Non-validators are assigned to ID zero.
+    /// @dev Validators have IDs greater than zero.
+    mapping(address => uint256) private _validatorId;
+
+    /// @notice Validator addresses indexed by ID.
+    /// @notice See the `validatorById` function.
+    /// @dev Invalid IDs are assigned to address zero.
```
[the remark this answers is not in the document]
guidanoli: I mean, I understand that we're not doing assignments as in Solidity statements, but I used the word "assign" as in English you'd say "I was assigned this task" or "He was assigned a unique number".

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -0,0 +1,152 @@
...
+
+    /// @notice Get the number of validators.
+    function numOfValidators() external view returns (uint256) {
+        return _numOfValidators;
+    }
+
+    /// @notice Get the ID of a validator.
+    /// @param validator The validator address
+    /// @dev Validators have IDs greater than zero.
+    /// @dev Non-validators are assigned to ID zero.
```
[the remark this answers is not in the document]
guidanoli: I think "to have ID as zero" sounds awkward.

onchain/rollups/contracts/consensus/quorum/Quorum.sol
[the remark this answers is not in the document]
guidanoli: Ok, maybe saying that "invalid IDs are assigned to address zero" may be misunderstood as "address zero is assigned to invalid IDs", while this might not be true! Address zero can be, theoretically, a validator and have a valid ID. So, saying that invalid IDs map to address zero is more correct.

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -0,0 +1,152 @@
...
+    function validatorId(address validator) external view returns (uint256) {
+        return _validatorId[validator];
+    }
+
+    /// @notice Get the address of a validator by its ID.
+    /// @param id The validator ID
+    /// @dev Validator IDs range from 1 to `N`, the total number of validators.
+    /// @dev Invalid IDs are assigned to the zero address.
```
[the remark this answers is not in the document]
guidanoli: I think "zero addresses" sounds awkward, as there is only one zero address.
How about "map to address zero"?

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -0,0 +1,152 @@
...
+    /// @param inFavorById The set of validators in favor of the claim
+    /// @dev `inFavorById` is a bitmap indexed by validator IDs.
+    struct Votes {
+        uint256 inFavorCount;
+        BitMaps.BitMap inFavorById;
+    }
+
+    /// @notice Votes indexed by claim.
+    /// @dev See the `numOfValidatorsInFavorOf` and `isValidatorInFavorOf` functions.
+    mapping(address => mapping(uint256 => mapping(uint256 => mapping(bytes32 => Votes))))
```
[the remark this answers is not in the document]
guidanoli: Ok, I will add a comment about how the keys of this nested mapping.

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -0,0 +1,152 @@
...
+contract Quorum is AbstractConsensus {
+    using BitMaps for BitMaps.BitMap;
+
+    /// @notice The total number of validators.
+    /// @notice See the `numOfValidators` function.
+    uint256 private immutable _numOfValidators;
+
+    /// @notice Validator IDs indexed by address.
+    /// @notice See the `validatorId` function.
+    /// @dev Non-validators are assigned to ID zero.
```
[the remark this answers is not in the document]
guidanoli: I think "have ID as zero" sounds awkward.

onchain/rollups/test/foundry/consensus/quorum/Quorum.t.sol
guidanoli: I will add a comment on this function to explain what it is testing exactly.

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -0,0 +1,152 @@
...
+        return _getVotes(application, r, epochHash).inFavorCount;
+    }
+
+    /// @notice Check whether a validator is in favor of a claim.
+    /// @param application The application address
+    /// @param r The input range
+    /// @param epochHash The epoch hash
+    /// @param id The ID of the validator
+    /// @return Whether validator is in favor of claim
+    /// @dev Assumes the provided ID is valid
```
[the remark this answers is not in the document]
ZzzzHui: Agree. That aligns with my "another angle" mentioned above, which makes sense to me as well :)

ZzzzHui (approved):

ZzzzHui (approved):

ZzzzHui:
rebased on top of main branch.
commit [fix: remove empty set check](https://github.com/cartesi/rollups-contracts/pull/49/commits/7ed3baae0fa6baec527f431ec1ef6b14f4e9ad20) and [fix: {first,last}InputIndex -> {first,last}Index](https://github.com/cartesi/rollups-contracts/pull/49/commits/f0837a7dbafaf483c181dd0fc0503d590e308c5c) are already in the main branch so those 2 commits are dropped.

ZzzzHui (approved):

pedroargento (approved):

Issue #55, opened 2023-08-21
optimize Quorum


ZzzzHui:
## 📚 Context

This issue is to optimize the [PR](https://github.com/cartesi/rollups-contracts/pull/49) to add quorum, which addresses [this issue](https://github.com/cartesi/rollups-contracts/issues/48). 

- We can optimize the Quorum by getting rid of AccessControlEnumerable and EnumerableSet. 
- Furthermore, instead of using `mapping(address => bool) voted;` to note who has voted, we could consider splitting a `uint256` into bits and utilize bit position to denote whether a validator has voted.

For details, please see [this comment](https://github.com/cartesi/rollups/pull/204#issuecomment-1682218456)

## ✔️ Solution

Since the last part (split a uint256) of the optimization can be complex and we are not sure whether we will go for that route, we will open 2 PR for this issue.

PR #57, opened 2023-08-30
feat: optimize quorum using bitmaps

ZzzzHui:


ZzzzHui:
After running `forge test --match-contract Quorum --gas-report --fuzz-seed 1`, here is the gas report:

| contracts/consensus/quorum/Quorum.sol:Quorum contract |                 |       |        |       |         |
|-------------------------------------------------------|-----------------|-------|--------|-------|---------|
| Deployment Cost                                       | Deployment Size |       |        |       |         |
| 1150219                                               | 6483            |       |        |       |         |
| Function Name                                         | min             | avg   | median | max   | # calls |
| getHistory                                            | 258             | 258   | 258    | 258   | 1       |
| submitClaim                                           | 4117            | 18907 | 4688   | 47917 | 3       |


Compared with the [benchmark](https://github.com/cartesi/rollups-contracts/pull/49#issuecomment-1730141760) (base branch `feature/quorum`)
- deployment cost reduced 613481 (34.78%)
- `submitClaim` function reduced: min 44512 (91.53%); avg 37879 (66.70%); max 24612 (33.93%)

The first validator to submit claim won't benefit from the optimization using Bitmaps, but the following validators using the same Bitmaps storage slot would. For the first storage slot, 255 validators are sharing; for each of the following slots, 256 are sharing.

Current test has only 3 validators, so we see a big discrepancy between `min` and `avg` gas cost/reduction for `submitClaim`. If we increase the number of validators to 255, the gas report becomes

| Function Name                                         | min             | avg  | median | max   | # calls |
|-------------------------------------------------------|-----------------|-------|--------|-------|---------|
| submitClaim                                           | 4117            | 4291 | 4117   | 47917 | 255     |

the `avg` gas cost is very close to `min`

---

Compared with the intermediary branch `feature/optimize-quorum`, in the case of 3 validators
- deployment cost  increased 3347 (0.29%)
- `submitClaim` function: `min` decreased 21658 (84.03%) `avg` decreased 14358 (43.16%) `max` increased 242 (0.50%)

guidanoli:
Here is the diff table in relation to your already optimized version (#55).

| Contract | Function | Min | Avg | Max |
| :- | :- | :-: | :-: | :-: |
| Quorum | | | | 3347 (0.29%) |
| Quorum | getHistory | 22 (9.32%) | 22 (9.32%) | 22 (9.32%) |
| Quorum | submitClaim | -21658 (-84.03%) | -14358 (-43.16%) | 242 (0.51%) |

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -70,10 +76,11 @@ contract Quorum is AbstractConsensus, PaymentSplitter {
     ///      its current history contract.
     function submitClaim(bytes calldata _claimData) external onlyValidator {
         Voted storage claimYeas = yeas[_claimData];
+        uint256 index = validatorIndex[msg.sender];
```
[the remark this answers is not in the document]
ZzzzHui: Agree. Since this is the only place using the modifier, we can remove the modifier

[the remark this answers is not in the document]
ZzzzHui: Changes applied and this change saves about 150 gas because it saves a second access to the same storage slot within the same transaction.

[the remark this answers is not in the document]
guidanoli: Yes, because the storage location gets "dirty" and a second read is cheaper.
But he can avoid the second read altogether. :grin:

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -68,12 +68,16 @@ contract Quorum is AbstractConsensus, PaymentSplitter {
     /// @dev Can only be called by a validator,
     ///      and the `Quorum` contract must have ownership over
     ///      its current history contract.
-    function submitClaim(bytes calldata _claimData) external onlyValidator {
+    function submitClaim(bytes calldata _claimData) external {
         Voted storage claimYeas = yeas[_claimData];
```
guidanoli, line 72: We should move this line after the if-clause, so that whenever someone (that is not a validator) calls this function, they waste less gas on a revert.

guidanoli:
We could abstract this logic of `index == 0` means `msg.sender` is not a validator in an `internal` function called `getValidatorIndex`, which raises `NotValidator` (instead of `OnlyValidator`).

ZzzzHui:
> We could abstract this logic of `index == 0` means `msg.sender` is not a validator in an `internal` function called `getValidatorIndex`, which raises `NotValidator` (instead of `OnlyValidator`).

I think `uint256 index = getValidatorIndex(msg.sender)` alone doesn't make it obvious that only validators can submit claims. If we want to make it obvious, we should either use comments or an internal function like
```
function onlyValidator(uint256 index) internal{
    if (index == 0) {
        revert NotValidator();
    }
}
function submitClaims{
    uint256 index = validatorIndex[msg.sender];
    onlyValidator(index);
    ....
}
```
which doesn't make a big difference from the current version. What do you think?

guidanoli:
> What do you think?

Maybe rename `onlyValidator` to `checkValidatorIndex`?

ZzzzHui:
Doesn't `onlyValidator` state its intention more clear than `checkValidatorIndex` :)

guidanoli:
> Doesn't onlyValidator state its intention more clear than checkValidatorIndex

Yes, but it's weird to have a function without an action verb.

ZzzzHui:
It's the same thing for modifier. We just modified the function from a modifier to an internal function for gas purpose. Name is the same. 
That being said, I think 
```
        if (index == 0) {
            revert OnlyValidator();
        }
```
is short and clear and only appears once. We could extract it to an internal function but that doesn't make it more readable, does it?

guidanoli:
> We could extract it to an internal function but that doesn't make it more readable, does it?

Ok, I think I agree with you. Let's leave it inline like this.

guidanoli:
> It's the same thing for modifier. We just modified the function from a modifier to an internal function for gas purpose.

Semantically, yes. But modifiers are different from functions. For me, at least, it's okay to call a modifier `onlyOwner` or `onlyRole` or `noReentrancy`, because it modifies the function. Now, the function itself does something, it calculates, creates, gets, sets, etc.

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -12,48 +13,47 @@ import {IHistory} from "../../history/IHistory.sol";
...
-        mapping(address => bool) voted;
+        // use BitMap to record who has voted
+        BitMaps.BitMap votedBitMap;
     }
     // Map a claim to struct Voted
     mapping(bytes => Voted) internal yeas;
 
     /// @notice Raised if not a validator
     error OnlyValidator();
-    modifier onlyValidator() {
```
xdaniortega, line 38: Hey! Just a question, why is it better to throw an error than just defining a modifier?

xdaniortega: I meant why did you change the modifier implementation in favor of de error?

guidanoli: Oh, I see what you mean now.
It was to avoid reading from the same storage slot twice.
See the thread in https://github.com/cartesi/rollups-contracts/pull/57#discussion_r1334915237.

xdaniortega: Good catch! Thanks

xdaniortega (approved):

guidanoli (approved):

PR #63, opened 2023-09-01
feat: improve dapp tests

ZzzzHui:


onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
```diff
@@ -28,16 +28,22 @@ contract EtherReceiver {
...
 contract CartesiDAppTest is TestBase {
     using LibServerManager for LibServerManager.RawFinishEpochResponse;
     using LibServerManager for LibServerManager.Proof;
 
+    // Outputs
+    // 0: notice 0xfafafafa
+    // 1: voucher ERC-20 transfer
+    // 2: voucher ETH withdrawal
+    // 3: voucher ERC-721 transfer
+    enum OutputCase {
```
guidanoli, line 40: I would rename this to `OutputName`.

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
```diff
@@ -28,16 +28,22 @@ contract EtherReceiver {
...
+        Notice,
```
guidanoli, line 41: I would rename this to `DummyNotice`.

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
```diff
@@ -28,16 +28,22 @@ contract EtherReceiver {
...
+        ERC20Voucher,
+        ETHVoucher,
+        ERC721Voucher
```
guidanoli, line 44: I would add the word `Transfer` after `ERC20` and `ERC721`, and the word `Withdrawal` after `ETH`.

guidanoli (commented):
Great job! I like that you renamed the `registerProof` function to `mockConsensus`, which is exactly what it is doing now.

Also, I think we should take advantage of Solidity type system, and change the signature of `getNotice`, `getVoucher`, `setupNoticeProof` and `setupVoucherProof` to accept the enumeration rather than the `uint256` index, and perform the conversion there.

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
```diff
@@ -667,10 +668,10 @@ contract CartesiDAppTest is TestBase {
         for (uint256 i; i < outputEnums.length; ++i) {
             LibServerManager.OutputEnum outputEnum = outputEnums[i];
             if (outputEnum == LibServerManager.OutputEnum.VOUCHER) {
-                Voucher memory voucher = getVoucher(i);
+                Voucher memory voucher = getVoucher(OutputName(i));
```
guidanoli, line 671: This part got a little weird. Maybe we should keep `getNotice` and `getVoucher` the way they were and create new `getNoticeByName` and `getVoucherByName` that do this conversion from enumeration to index and call internally `getNotice` and `getVoucher` respectively.

guidanoli: What I find it weird is that, in this function, we are converting `uint256` into `OutputName`, and then in `getVoucher`, we are converting it back into `uint256`.

ZzzzHui: Agreed and addressed :)

guidanoli (approved):

PR #67, opened 2023-09-04
feat: make `CartesiDApp.t.sol` remove any extra input JSON files

### LINKED ISSUE #61: Remove extra input JSON files
## 📚 Context

When you're testing things out, and realize you created more inputs than you actually needed, you might delete these extra inputs in `CartesiDApp.t.sol`. The JSON files, however, are still on the `onchain/rollups/test/foundry/dapp/helper/input` folder, and so they will still get processed by the machine.

## ✔️ Solution

We should delete any extra input JSON file to keep the machine from processing any unnecessary input. The algorithm is simple. After `writeInputs`, we can call a new function `removeExtraInputs` that looks for a continuous sequence of inputs starting from index `outputEnums.length`. For each file found, call [`vm.removeFile`](https://book.getfoundry.sh/cheatcodes/fs).

guidanoli:
Closes #61

### DIFF
--- a/onchain/rollups/foundry.toml
+++ b/onchain/rollups/foundry.toml
@@ -7,7 +7,7 @@ cache_path  = 'forge-cache'
 solc_version = '0.8.19'
 fs_permissions = [
     { access = "read", path = "./test/foundry/dapp/helper/output/"},
-    { access = "write", path = "./test/foundry/dapp/helper/input/"}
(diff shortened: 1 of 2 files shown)


ZzzzHui (approved):

guidanoli:
Rebased.

ZzzzHui (approved):

Issue #76, opened 2023-09-08
Raise custom error when `getVoucher` or `getNotice` on an output that is not of the expected type


guidanoli:
## 📚 Context

Currently, we are asserting that the output being retrieved is the correct type.
This, however, gives little clue to the developer of what went wrong.

## ✔️ Solution

A better solution would be to raise a custom error saying that the function expected the output to be a voucher (or a notice), but it wasn't.

PR #78, opened 2023-09-12
feat: add extra checks in `CartesiDApp.t.sol`

### LINKED ISSUE #74: Add bound check to `CartesiDApp.t.sol`
## 📚 Context

If you call `getVoucher` or `getNotice` with a high enough value, it will revert with an `Index out of bounds` error message. This might be due to the fact that the user didn't run `yarn proofs:update` on a branch that had more inputs than in the `finish_epoch_response.json`.

## ✔️ Solution

Check if the input index is greater or equal to `outputEnums.length`, and raise a nice error message if so.

guidanoli:
Closes #74 and #76


ZzzzHui (commented):
I'm think whether we shouold rename some of these "inputIndex" to something like "inputIndexWithinEpoch" or "caseIndex". Although it also make sense that "inputIndex" can refer to index of those inputs for docker, in other places we use "inputindex" refer to index in the input box.

BTW the CI tests didn't pass.

xdaniortega (approved):
👍🏾

PR #85, opened 2023-09-19
fix: correct the use of inputIndex in tests

ZzzzHui:
- correct the naming of `inputIndex` to `inputIndexWithinEpoch` if it's used incorrectly
- ~~abbreviate `inputIndexWithinEpoch` as `IIWE` if it's compounded with other words~~

### DIFF
--- a/onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
+++ b/onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
@@ -42,14 +42,17 @@ contract CartesiDAppTest is TestBase {
     error UnexpectedOutputEnum(
         LibServerManager.OutputEnum expected,
         LibServerManager.OutputEnum obtained,
-        uint256 inputIndex
+        uint256 inputIndexWithinEpoch
     );
 
-    error InputIndexOutOfBounds(uint256 length, uint256 inputIndex);
+    error InputIndexWithinEpochOutOfBounds(
+        uint256 length,
+        uint256 inputIndexWithinEpoch
+    );
 
     error ProofNotFound(
         LibServerManager.OutputEnum outputEnum,
-        uint256 inputIndex
+        uint256 inputIndexWithinEpoch
     );
 
     CartesiDApp dapp;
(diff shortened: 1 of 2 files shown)


onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
```diff
@@ -42,14 +42,15 @@ contract CartesiDAppTest is TestBase {
     error UnexpectedOutputEnum(
         LibServerManager.OutputEnum expected,
         LibServerManager.OutputEnum obtained,
-        uint256 inputIndex
+        uint256 inputIndexWithinEpoch
     );
 
-    error InputIndexOutOfBounds(uint256 length, uint256 inputIndex);
+    // IIWE stands for InputIndexWithinEpoch
+    error IIWEOutOfBounds(uint256 length, uint256 inputIndexWithinEpoch);
```
guidanoli, line 49: If a developer (not too familiar with the code) runs `forge test` and gets this error, they will have no idea what `IIWE` means. I would, therefore, not abbreviate this error name.
```suggestion
    error InputIndexWithinEpochOutOfBounds(uint256 length, uint256 inputIndexWithinEpoch);
```

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
```diff
@@ -622,29 +623,39 @@ contract CartesiDAppTest is TestBase {
         vouchers[index] = Voucher(destination, payload);
     }
 
-    function checkInputIndex(uint256 inputIndex) internal view {
+    // IIWE stands for InputIndexWithinEpoch
+    function checkIIWE(uint256 inputIndexWithinEpoch) internal view {
```
guidanoli, line 627: This might save some characters, but we lose clarity by abbreviating `InputIndexWithinEpoch` to `IIWE`.
I would, therefore, avoid abbreviating it.
This applies to `IIWEStr` as well.

onchain/rollups/test/foundry/util/LibServerManager.sol
```diff
@@ -33,7 +33,7 @@ library LibServerManager {
 
     struct RawProof {
         bytes32 context;
-        string inputIndex;
+        string inputIndexWithinEpoch;
```
guidanoli, line 36: If you take a peek in `finish_epoch_response.json`, you'll see this is indeed `inputIndex`.

ZzzzHui: Then we should modify the `proves` function in the `LibServerManager.sol`. Instead of  checking `Proof.inputIndex`, we will be checking `Proof.validity.inputIndexWithinEpoch`.

ZzzzHui:
BTW, is there a difference between `outputIndex` and `outputIndexWithinInput`? I saw these 2 names but they seem to mean the same thing?

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
ZzzzHui: Yeah. It's always a tradeoff between being clear and concise. Since they don't appear often, we could use the lengthy form

onchain/rollups/test/foundry/util/LibServerManager.sol
guidanoli: Yes, this makes sense.

guidanoli:
> BTW, is there a difference between `outputIndex` and `outputIndexWithinInput`? Is `outputIndex` global like `inputIndex` is?

`outputIndexWithinInput` lives inside the `validity` structure, which could change. Meanwhile, `inputIndex` and `outputIndex` are in the same level as `validity` and aim to help filter the correct proof.

ZzzzHui:
> outputIndexWithinInput lives inside the validity structure, which could change

I don't get this part. What change do you mean :)

Apart from being in different levels, is there a difference between `outputIndex` and `outputIndexWithinInput`? In other words, should they be 2 different names

guidanoli (approved):

guidanoli:
> Apart from being in different levels, is there a difference between outputIndex and outputIndexWithinInput?

I believe they contain the same data.
@gligneul Can you confirm?

Issue #2, opened 2023-08-17
Rename fields in `InputAdded` event.


guidanoli:
## 📚 Context

The `IInputBox` interface specifies an event called `InputAdded`.
We noticed that we could improve the naming of some of its arguments.

## ✔️ Solution

* Rename `inputIndex` --> `index`
* Rename `input` --> `payload`

guidanoli:
FYI @miltonjonat

Issue #84, opened 2023-09-15
Blockchain-agnostic Inputs


guidanoli:
## 📚 Context

Currently, an input is composed of two parts: the input data and the input metadata. The latter includes: the block number and timestamp, the address of whoever sent the input, and the index of the input in the input box.

This, however, exposes to the Cartesi Machine information that is completely EVM-specific, while the Cartesi Machine should be blockchain-agnostic.

## ✔️ Solution

Following the zen of #42, we would like inputs to also be self-contained blobs.

Input payloads are already blobs. So, what we would like to do is ABI-encode the input metadata along with the payload and call it the input blob itself. The hash of this blob will, therefore, constitute as the input hash, which will play a role in arbitration.

Looking far into the future, the Cartesi Machine could be fed with inputs from blockchains other than Ethereum, or we might want to change the blockchain metadata schema. We should, therefore, be able to encode different blockchain metadata and know how to decode it just from a simple inspection of the input blob.

Such type of encoding is possible with the `abi.encodeCall` function, which will also be used to encode outputs. Decoding inputs that are encoded in this way should be straightforward, given that you can identify the input format just by looking at the first four bytes of the blob, which constitutes the function selector, in Solidity terms.

## :memo: Specification

The input metadata is currently composed of the following fields:

- `address sender`: the address of the input sender
- `uint256 blockNumber`: The number of the block in which the input was added
- `uint256 blockTimestamp`: The timestamp of the block in which the input was added (in UNIX epoch format)
- `uint256 index`: The index of the input (in relation to all inputs ever sent to the DApp)

If we are going to use `abi.encodeCall`, we need to decide on the function name that clearly describes the type of input blob. Since this input is coming from an EVM-compatible blockchain, and will trigger an advance request for the off-chain machine, we might call it `EvmAdvance` or `EVMAdvance` (depending on whether you like to keep acronyms capitalized in `PascalCase`). I would generally go for `EVMAdvance`, but I think `EvmAdvance` looks better. The encoding would look something like the following.

```solidity
function EvmAdvance(
    address sender,
    uint256 blockNumber,
    uint256 blockTimestamp,
    uint256 index,
    bytes calldata payload
) external;
```

ZzzzHui:
[...]

PR #89, opened 2023-09-25
feat: blockchain agnostic inputs

ZzzzHui:
Not sure if this draft is somewhat close to the expected solution? Please let me know if it's not :)


guidanoli:
Almost there. The algorithm is actually this:

1. Encode the blob using `abi.encodeWithSignature`.
2. Check if the blob size exceeds the limit.
3. Return the hash of the blob.

ZzzzHui:
> Almost there. The algorithm is actually this:
> 
> 1. Encode the blob using `abi.encodeWithSignature`.
> 2. Check if the size of the blob exceeds the limit.
> 3. Return the hash of the blob.

These were implemented in the LibInput, right

guidanoli:
> These were implemented in the LibInput, right

These should be implemented in `LibInput`.

ZzzzHui:
I meant they were already implemented, right?

guidanoli:
Not quite. We won't be hashing the metadata and payload separately. Instead, we'll ABI-encoded the metadata and payload, and call this the input, which is then hashed.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `abi` directory to the `@cartesi/rollups` package, ideal for language bindings.
 
+### Changed
+
+- Changed inputs encodings to be blockchain agnostic
```
guidanoli, line 16: Actually, this would go to 2.0.0, and the upcoming release is 1.1.0...
That's why I think we should use change sets, to have fewer problems while recording changes in each PR.

onchain/rollups/contracts/inputs/IInputBox.sol
```diff
@@ -9,7 +9,7 @@ interface IInputBox {
     /// @param dapp The address of the DApp
     /// @param inputIndex The index of the input in the input box
     /// @param sender The address that sent the input
-    /// @param input The contents of the input
+    /// @param input The input payload
```
guidanoli, line 12: Maybe fix #2 in this PR as well?

onchain/rollups/test/foundry/inputs/InputBox.t.sol
```diff
@@ -158,13 +158,33 @@ contract InputBoxTest is Test {
...
+            0,
+            0,
+            new bytes(32)
+        );
+        // number of bytes in input blob excluding input payload
+        uint256 extraBytes = blob.length - 32;
+        // because it's abi encoded, input payloads are stored as multiples of 32 bytes
+        uint256 maxLength = ((CanonicalMachine.INPUT_MAX_SIZE - extraBytes) /
+            32) * 32;
+        return maxLength;
```
guidanoli, line 176: Why not return this expression right away, instead of storing this in a variable and returning it. Then, maybe it all fits in one line.

onchain/rollups/test/foundry/inputs/InputBox.t.sol
```diff
@@ -158,13 +158,33 @@ contract InputBoxTest is Test {
         assertEq(inputBox.getNumberOfInputs(_dapp), 0);
     }
 
+    function getMaxInputPayloadLength() internal pure returns (uint256) {
+        // an input blob instance
```
guidanoli, line 162: I feel like these comments only add noise, since the code is pretty straightforward and easy to understand.

CHANGELOG.md
ZzzzHui: Yeah. We could start using changeset

onchain/rollups/test/foundry/inputs/InputBox.t.sol
ZzzzHui: I don't see this add noise actually

onchain/rollups/contracts/library/LibInput.sol
guidanoli, line 1: Maybe create an `encodeEvmInput` function to help in the tests?

onchain/.changeset/mean-news-float.md
```diff
@@ -0,0 +1,6 @@
+---
+"@cartesi/rollups": major
+---
+
+Now inputs are blockchain-agnostic and self-contained blobs. EVM inputs will be encoded with signature of evm.
```
guidanoli, line 5: First change.

```suggestion
Inputs are now blockchain-agnostic and self-contained blobs. For example, inputs added by EVM contracts like `InputBox` are encoded using `abi.encodeWithSignature` and contain EVM-specific metadata like `msg.sender` and `block.timestamp`.
```

onchain/.changeset/mean-news-float.md
```diff
@@ -0,0 +1,6 @@
...
+Also rename fields in the InputAdded event. inputIndex --> index and input --> payload.
```
guidanoli, line 6: Second change.

```suggestion
Fields in the `InputAdded` event were renamed. `inputIndex` was renamed to `index`, and `input` was renamed to `payload`.
```

onchain/rollups/contracts/inputs/InputBox.sol
```diff
@@ -8,11 +8,11 @@ import {LibInput} from "../library/LibInput.sol";
 
 /// @title Input Box
 ///
-/// @notice Trustless and permissionless contract that receives arbitrary blobs
-/// (called "inputs") from anyone and adds a compound hash to an append-only list
+/// @notice Trustless and permissionless contract that receives arbitrary
+/// input payload from anyone and adds a compound hash to an append-only list
```
guidanoli, line 12:
```suggestion
/// data from anyone and adds a compound hash to an append-only list
```

onchain/rollups/contracts/inputs/InputBox.sol
```diff
@@ -8,11 +8,11 @@ import {LibInput} from "../library/LibInput.sol";
...
 /// (called "input box"). Each DApp has its own input box.
 ///
-/// The hash that is stored on-chain is composed by the hash of the input blob,
+/// The hash that is stored on-chain is composed by the hash of the input payload,
 /// the block number and timestamp, the input sender address, and the input index.
```
guidanoli, line 16:
```suggestion
/// The input hash is composed by the input payload, the block number and timestamp,
/// the address of the input sender, and the index of the input.
```

onchain/rollups/contracts/library/LibInput.sol
```diff
@@ -12,36 +12,57 @@ library LibInput {
...
-        uint256 inputIndex
-    ) internal pure returns (bytes32) {
-        if (input.length > CanonicalMachine.INPUT_MAX_SIZE) {
-            revert InputSizeExceedsLimit();
-        }
-
-        bytes32 keccakMetadata = keccak256(
-            abi.encode(
+        uint256 index,
+        bytes memory payload
```
guidanoli, line 27: Please make this `calldata`, for gas cost reasons. In the tests, you can create a contract that receives `calldata` which you can call with a payload on `memory`.

onchain/rollups/contracts/library/LibInput.sol
```diff
@@ -12,36 +12,57 @@ library LibInput {
     /// @notice Raised when input is larger than the machine limit.
     error InputSizeExceedsLimit();
 
-    /// @notice Summarize input data in a single hash.
+    /// @notice Encode input with EvmInput signature
```
guidanoli, line 15: This is too much detail.

```suggestion
    /// @notice Encode an EVM input
```

onchain/rollups/test/foundry/inputs/InputBox.t.sol
```diff
@@ -158,28 +158,45 @@ contract InputBoxTest is Test {
...
+        bytes memory blob = LibInput.encodeEvmInput(
```
guidanoli, line 163: Here, you can call a contract that has a function with the exact same signature of `encodeEvmInput` which calls this function from `LibInput`. It's perfectly ok to call an external function that expects `calldata` with `memory`.

guidanoli:
We can make this PR has two commits.
- One that solves #84 + changeset
- One that solves #2 + changeset

onchain/rollups/contracts/library/LibInput.sol
ZzzzHui: I'm curious, how would other contracts (besides the test contract) actually call this function then? Do they also create an intermediary contract or though low level calls to call this function?

onchain/rollups/test/foundry/inputs/InputBox.t.sol
```diff
@@ -254,3 +271,22 @@ contract InputBoxTest is Test {
         assertEq(sum, totalNumOfInputs, "total number of inputs");
     }
 }
+
+contract LibInputCaller {
```
guidanoli, line 275: You don't have to create a new contract for this. You can define in the test contract itself and call with an extra `this.` before, like here:

- Call: https://github.com/cartesi/rollups-contracts/blob/a61d5f736b208ad635a9b9ecf7ea086621a31ae8/onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol#L723
- Definition: https://github.com/cartesi/rollups-contracts/blob/a61d5f736b208ad635a9b9ecf7ea086621a31ae8/onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol#L705-L707

onchain/rollups/contracts/library/LibInput.sol
```diff
@@ -12,36 +12,57 @@ library LibInput {
...
-            revert InputSizeExceedsLimit();
-        }
-
-        bytes32 keccakMetadata = keccak256(
-            abi.encode(
+        uint256 index,
+        bytes calldata payload
+    ) internal pure returns (bytes memory) {
+        return
+            abi.encodeWithSignature(
```
guidanoli, line 30: So, recently I found a nice way to encode function calls in a nice type-safe way. See my comment on the output unification issue [here](https://github.com/cartesi/rollups-contracts/issues/42#issuecomment-1791371612). I was thinking if we could create an `Inputs.sol` interface on the `contracts/common` directory and define the `EvmInput` function there. What do you think?

This should also simplify the tests, since they will be able to more easily encode EVM inputs with `abi.encodeCall` as well. No need for an auxiliary function that calls the library function, etc.

As a result, we can encode the input directly on the `InputBox` contract. We can also check there if the input is smaller than the maximum input size, and hash it. With these changes, we could delete the `LibInput.sol` file altogether.

onchain/.changeset/cuddly-ears-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cartesi/rollups": minor
+---
+
+Fields in the `InputAdded` event were renamed. `inputIndex` was renamed to `index`, and `input` was renamed to `payload`.
```
guidanoli, line 5: I know I called them fields, but after reading the Solidity documentation, it seems that they are normally called arguments/parameters. The distinction is the same with functions: arguments are values (like `123` in `f(123)`), while parameters are variables (like `x` in `function f(x) end`).

Also, it seems that the right way to say is "renamed as".

```suggestion
Two parameters of the `InputAdded` event were renamed: `inputIndex` was renamed as `index`, and `input` was renamed as `payload`.
```

onchain/.changeset/mean-news-float.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cartesi/rollups": major
+---
+
+Inputs are now blockchain-agnostic and self-contained blobs. For example, inputs added by EVM contracts like `InputBox` are encoded using `abi.encodeWithSignature` and contain EVM-specific metadata like `msg.sender` and `block.timestamp`.
```
guidanoli, line 5: The fact that inputs are encoded with `abi.encodeWithSignature` is a mere implementation detail. Also, we might use `abi.encodeCall` instead for type safety reasons. Either way, let's focus on the fact that inputs now contain metadata.

```suggestion
Inputs are now blockchain-agnostic and self-contained blobs. For example, inputs added by EVM contracts like `InputBox` contain EVM-specific metadata like `msg.sender` and `block.timestamp`.
```

onchain/rollups/contracts/library/LibInput.sol
ZzzzHui: Good idea Gui 👍 

>  we can encode the input directly on the InputBox contract

Yeah the `computeEvmInputHash` library function is quite short that we can directly write it in the `InputBox` contract.

onchain/rollups/contracts/inputs/IInputBox.sol
```diff
@@ -7,26 +7,26 @@ pragma solidity ^0.8.8;
...
+    /// @param index The index of the input in the input box
     /// @param sender The address that sent the input
-    /// @param input The contents of the input
+    /// @param payload The input payload
     /// @dev MUST be triggered on a successful call to `addInput`.
     event InputAdded(
         address indexed dapp,
-        uint256 indexed inputIndex,
+        uint256 indexed index,
         address sender,
```
guidanoli, line 17: We don't *need* to emit the input sender as argument to the `InputAdded` event anymore, as this information is now encoded in the input itself, and is only really useful to the DApp back-end (while `dapp` and `index` are still useful for the node to process inputs to a DApp and filter/order them by index).

onchain/rollups/contracts/inputs/IInputBox.sol
```diff
@@ -7,26 +7,26 @@ pragma solidity ^0.8.8;
...
-        bytes input
+        bytes payload
```
guidanoli, line 18: Maybe it's wise to keep this parameter being name `input`, because it is no longer the `payload` provided to the `addInput` function. Also, we may need to make this distinction clear in the documentation of the `IInputBox` interface.

onchain/rollups/contracts/inputs/IInputBox.sol
```diff
@@ -7,26 +7,26 @@ pragma solidity ^0.8.8;
...
     );
 
     /// @notice Add an input to a DApp's input box.
     /// @param _dapp The address of the DApp
-    /// @param _input The contents of the input
+    /// @param _payload The input payload
     /// @return The hash of the input plus some extra metadata
     /// @dev MUST fire an `InputAdded` event accordingly.
     ///      Input larger than machine limit will raise `InputSizeExceedsLimit` error.
```
guidanoli, line 26: We can move the `InputSizeExceedsLimit` error to the interface, now that we are getting rid of the `LibInput` library.

onchain/rollups/contracts/inputs/IInputBox.sol
ZzzzHui: I think it's more gas efficient to emit the payload rather than the whole input blob. Because `index` is already emitted and `block.number`,`block.timestamp` are retrievable as event meta data  too.

onchain/rollups/contracts/inputs/IInputBox.sol
ZzzzHui: We do need if we only emit payload rather than the whole input blob

guidanoli: True. Maybe we need to discuss this with the node team.

onchain/rollups/contracts/inputs/IInputBox.sol
guidanoli: True, let's discuss this with the node team.

ZzzzHui: I see that @gligneul agreed with the event format as (`dapp`,`index`,`input`). I will make the changes

pedroargento (dismissed):

ZzzzHui:
After taking a closer look, the CI failed on the same reason that was fixed [in the main branch](https://github.com/cartesi/rollups-contracts/pull/124). So after merging it into main, the failure should resolve on its own. 
In the previous PR, the error message showed up as `Assertion failed` but in this PR it showed up as `<no data>` so I thought they are different errors. But after taking a deeper look, they failed on the same thing.

guidanoli:
> After taking a closer look, the CI failed on the same reason that was fixed [in the main branch](https://github.com/cartesi/rollups-contracts/pull/124). So after merging it into main, the failure should resolve on its own. In the previous PR, the error message showed up as `Assertion failed` but in this PR it showed up as `<no data>` so I thought they are different errors. But after taking a deeper look, they failed on the same thing.

These invariant errors do give us more headaches than guarantees...

onchain/rollups/test/foundry/util/TestBase.sol
```diff
@@ -17,4 +19,16 @@ contract TestBase is Test {
         vm.assume(addr != MULTICALL3_ADDRESS);
         _;
     }
+
+    function encodeEvmAdvance(
```
guidanoli, line 23: I don't think this is the best place for this function.
You can create a separate helper library under `test/foundry/util` just for this task, if you want. But this contract is inherited by many test contracts, many of which have nothing to do with inputs.

guidanoli:
Please rebase this on top of `main`. :-)

onchain/rollups/contracts/common/Inputs.sol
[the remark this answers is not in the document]
gligneul: It isn't necessary because the cartesi machine expects an inspect to be just the payload. Hence, adding the inspect to the interface would be confusing since it isn't used.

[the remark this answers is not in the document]
guidanoli: But wasn't @mpolitzer going to encode both advances and inspects with Solidity ABI, in [his PR](https://github.com/cartesi/machine-emulator-tools/pull/24)?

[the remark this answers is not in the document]
mpolitzer: The plan is to ABI encode both `advance` and `inspect`.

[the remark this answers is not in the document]
guidanoli: Okay, so... should the node encode inputs to the machine (be it an advance or an inspect request), would it be useful for us to define the inspect signature as well?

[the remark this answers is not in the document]
gligneul: Yes, makes sense.

guidanoli:
We should update the `EvmAdvance` signature to consider #144 and #145. See https://github.com/cartesi/rollups-contracts/issues/84#issuecomment-1818081690.

ZzzzHui:
> > After taking a closer look, the CI failed on the same reason that was fixed [in the main branch](https://github.com/cartesi/rollups-contracts/pull/124). So after merging it into main, the failure should resolve on its own. In the previous PR, the error message showed up as `Assertion failed` but in this PR it showed up as `<no data>` so I thought they are different errors. But after taking a deeper look, they failed on the same thing.
> 
> These invariant tests give us more headaches than guarantees...

I did have a bit of similar feelings as well

onchain/rollups/.changeset/cuddly-ears-obey.md
```diff
@@ -0,0 +1,7 @@
+---
+"@cartesi/rollups": major
+---
+
+`inputIndex` is renamed as `index`
+`input` is renamed as `payload` and `input` now refers to the whole input blob
+event `InputAdded` now contains `dapp`, `index` and `input`
```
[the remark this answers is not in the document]
ZzzzHui: Don't you think that's quite a lot of details in a changeset/changelog file for one commit?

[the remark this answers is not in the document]
ZzzzHui: And do you mean to have 2 changeset files for one commit or split the commit into 2 commits as well?

ZzzzHui:
> We should update the `EvmAdvance` signature to consider #144 and #145. See [#84 (comment)](https://github.com/cartesi/rollups-contracts/issues/84#issuecomment-1818081690).

The amount of discussions in this PR is already quite large that I would like to address this in a new PR :)

ZzzzHui:
> Please rebase this on top of `main`. :-)

Will rebase after https://github.com/cartesi/rollups-contracts/pull/136#pullrequestreview-1737661313 being merged into main branch, because the current PR is based on a similar commit

guidanoli:
Ok, #136 has been merged.

onchain/rollups/test/foundry/inputs/InputBox.t.sol
```diff
@@ -22,60 +24,82 @@ contract InputBoxTest is Test {
 
     function testAddLargeInput() public {
         address app = vm.addr(1);
-        uint256 max = CanonicalMachine.INPUT_PAYLOAD_MAX_SIZE;
+        uint256 max = _getMaxInputPayloadLength();
 
         _inputBox.addInput(app, new bytes(max));
 
+        uint256 largeLength = EvmAdvanceEncoder
+            .encode(address(this), 1, new bytes(max + 1))
```
guidanoli, line 32: You can store the `new bytes(max + 1)` value in a variable called `payload`, since you'll be passing that to `addInput` later anyway.

onchain/rollups/test/foundry/inputs/InputBox.t.sol
```diff
@@ -22,60 +24,82 @@ contract InputBoxTest is Test {
...
+                )
             );
             // test if input hash is the same as in InputBox
             assertEq(inputHash, _inputBox.getInputHash(app, i));
             // test if input hash is the same as returned from calling addInput() function
             assertEq(inputHash, returnedValues[i]);
         }
     }
+
+    function _getMaxInputPayloadLength() internal pure returns (uint256) {
```
guidanoli, line 95: Why not store this constant value in an immutable variable in this contract called `MAX_INPUT_PAYLOAD_LENGTH`, and use it instead? Then, you can move the code of function to the constructor of the test contract.

onchain/rollups/contracts/common/Inputs.sol
```diff
@@ -0,0 +1,22 @@
...
+/// @title Inputs
+/// @notice Defines the signatures of inputs.
+interface Inputs {
+    /// @notice An advance request from an EVM-compatible blockchain to a Cartesi Machine.
+    /// @param sender The address of whoever sent the input
+    /// @param blockNumber The number of the block in which the input was added
+    /// @param blockTimestamp The timestamp of the block in which the input was added
+    /// @param index The input index
+    /// @param payload The payload provided by the sender
+    function EvmAdvance(
```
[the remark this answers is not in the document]
guidanoli: We are not going to use it, but it's going to be there as a reference for other teams.
It's like the signature for `Notice`. It's not going to be used in our contracts, but elsewhere, namely in the `libcmt` library.

onchain/rollups/contracts/common/CanonicalMachine.sol
```diff
@@ -20,9 +20,9 @@ library CanonicalMachine {
     /// @notice Keccak-256 output size (32 bytes).
     Log2Size constant KECCAK_LOG2_SIZE = Log2Size.wrap(5);
 
-    /// @notice Maximum input payload size (~2 megabytes).
+    /// @notice Maximum input size (~2 megabytes).
     /// @dev The offset and size fields use up the extra 64 bytes.
-    uint256 constant INPUT_PAYLOAD_MAX_SIZE = (1 << 21) - 64;
+    uint256 constant INPUT_MAX_SIZE = (1 << 21) - 64;
```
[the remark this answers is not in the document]
guidanoli: From the size and offset fields added by the `abi.encode`.

onchain/rollups/test/foundry/inputs/InputBox.t.sol
ZzzzHui: I think it's cleaner as a function alone

onchain/rollups/contracts/common/CanonicalMachine.sol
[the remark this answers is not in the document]
ZzzzHui: Is this the max size for the input blob or just input payload?
If for the blob, we don't need to `-64`;
If for the payload, shouldn't we consider the selector and `msg.sender, block.number, block.timestamp, index` as well?
I probably have missed something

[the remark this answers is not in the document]
guidanoli: This is the maximum size for the input blob.
Has something changed in this regard, @mpolitzer?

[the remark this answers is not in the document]
ZzzzHui: If it's for the blob, we shouldn't `-64`, right?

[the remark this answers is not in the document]
guidanoli: From what I understood from talking to @mpolitzer, the input blob is still `abi.encoding` before being written to the buffer in memory, thus the 64 bytes discount.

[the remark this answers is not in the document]
ZzzzHui: Here is my breakdown of the input blob:
|                 | # bytes |
|-----------------|---------|
| selector        | 4       |
| msg.sender      | 32      |
| block.number    | 32      |
| block.timestamp | 32      |
| index           | 32      |
| offset          | 32      |
| length          | 32      |
| payload         | DYNAMIC      |

By `INPUT_MAX_SIZE`, are we referring to the max size of the same structure above?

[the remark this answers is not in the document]
ZzzzHui: Maybe it's easier to discuss by referring to codes. For example, [this line](https://github.com/cartesi/rollups-contracts/blob/77e0df4bba450df6b4fb9764a3fb6394781aa56d/onchain/rollups/contracts/inputs/InputBox.sol#L28) would become 
```
if (input.length - 64 > CanonicalMachine.INPUT_MAX_SIZE) {
```
if we take out `offset` and `length`. 
I don't understand why subtract 64 from both sides of the equation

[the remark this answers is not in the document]
guidanoli: This is getting confusing. Let's schedule a meeting with @mpolitzer. :-)

onchain/rollups/.changeset/grumpy-coins-drop.md
```diff
@@ -0,0 +1,8 @@
+---
+"@cartesi/rollups": major
+---
+
+-   Inputs are now blockchain-agnostic and self-contained blobs.
+-   Removed `sender` from `InputAdded` event.
+-   Library `LibInput` is removed.
+-   Renamed `input` as `payload`
```
guidanoli, line 8: The first line can be in a dedicated changeset file.
The removal of the `LibInput` can be moved to the `seven-teachers-behave.md` changeset file.
The rest of the changes can be nested in a "Modified the `IInputBox` interface:" changeset file.
Also, the "Renamed `input` as `payload`" change should be instantiated for every entity in the `IInputBox` interface that was affected.

guidanoli (approved):

onchain/rollups/.changeset/six-doors-grow.md
```diff
@@ -0,0 +1,8 @@
+---
+"@cartesi/rollups": major
+---
+
+Modified the IInputBox interface:
+
+-   removed `sender` from `InputAdded` event.
+-   correct the naming of `input` and `payload`.
```
guidanoli, line 8: Renaming is not very important for positional parameters. Solidity doesn't support keyword arguments like Python, so we don't have to worry. They are backwards-compatible. Meanwhile, semantic changes have to be notified.

```suggestion
-   Modified the `InputAdded` event:

    -   Removed the `sender` parameter.
    -   Changed the semantics of the `input` parameter.

-   Added an `InputTooLarge` error. 
```

onchain/rollups/.changeset/six-doors-grow.md
```diff
@@ -0,0 +1,10 @@
+---
+"@cartesi/rollups": major
+---
+
+-   Modified the `InputAdded` event:
+
+    -   Removed the `sender` parameter.
+    -   Changed the semantics of the `input` parameter.
+
+-   Added an `InputTooLarge` error.
```
guidanoli, line 10: Please keep the line you wrote, before all of these list items.

```md
Modified the `IInputBox` interface:
```

guidanoli (approved):

PR #107, opened 2023-10-09
Propagate errors on voucher execution in CartesiDApps

xdaniortega:


xdaniortega:
This approach does not handle any returned value at `executeVoucher` function as errors now are bubbled up and not reverting means the execution is always sucessfull.

xdaniortega:
If you find this approach correct I will catch correctly expected reverts with `vm.expectRevert();`

onchain/rollups/contracts/dapp/CartesiDApp.sol
```diff
@@ -133,16 +134,14 @@ contract CartesiDApp is
...
+        Address.functionCall(_destination, _payload);
 
-        // if properly executed, mark it as executed and emit event
-        if (succ) {
-            voucherBitmask.setBit(voucherPosition, true);
-            emit VoucherExecuted(voucherPosition);
-        }
-
-        return succ;
+        // Succesfull execution is ensured previously, mark it as executed and emit event
```
guidanoli, line 142: We already said in the previous line that failure is handled internally, so clearly program flow will only reach this line if the voucher was successfully executed.

```suggestion
        // mark it as executed and emit event
```

onchain/rollups/contracts/dapp/CartesiDApp.sol
```diff
@@ -133,16 +134,14 @@ contract CartesiDApp is
             revert VoucherReexecutionNotAllowed();
         }
 
-        // execute voucher
-        (bool succ, ) = _destination.call(_payload);
+        // execute voucher, handles success and failure internally.
+        // If `_destination` reverts with a revert reason or custom
+        // error, it is bubbled up by this function.
```
guidanoli, line 139: Revert reason and custom error are Solidity concepts, but we might be calling contracts not written in Solidity. To adopt a more language-agnostic view, we could simply say:

```suggestion
        // execute voucher
        // propagates any error in case of failure
```

Also, it might be worth to mention that, if the low-level call succeeds, we are discarding the return data entirely. I would like, however, on another PR, to make this function to forward this return data as-is.

onchain/rollups/contracts/dapp/ICartesiDApp.sol
```diff
@@ -46,14 +46,14 @@ interface ICartesiDApp {
     /// @param _payload The payload, which—in the case of Solidity contracts—encodes a function call
     /// @param _proof The proof used to validate the voucher against
     ///               a claim submitted by the current consensus contract
-    /// @return Whether the execution was successful or not
-    /// @dev On a successful execution, emits a `VoucherExecuted` event.
+    /// @dev No returned value, instead reverts on failure.
+    ///      On a successful execution, emits a `VoucherExecuted` event.
```
guidanoli, line 50: We don't need to say that a function returns no value, as this information is already embedded in the signature. We should say, however, that the error is propagated as-is.

```suggestion
    /// @dev On a successful execution, emits a `VoucherExecuted` event.
    ///      On a failed execution, propagates the error data as-is.
```

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
guidanoli, line 1: Here, we should avoid doing just `vm.expectRevert()` as it will allow any error whatsoever to pass the test. Instead, we'd like to be more specific in what error we expect to be raised. You may pass a string, an ABI-encoded custom error, or just the error selector. Check the [documentation](https://book.getfoundry.sh/cheatcodes/expect-revert).

guidanoli:
> If you find this approach correct I will catch correctly expected reverts with vm.expectRevert();

Yes, this is the intended approach. :-)

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
guidanoli: Sorry, just saw your comment. So, you are aware. :-)

onchain/rollups/contracts/dapp/CartesiDApp.sol
guidanoli, line 1: Also, you may use `Address` as a library to handle addresses, like so:

```solidity
using Address for address;
```

Then you may use the functions from `Address` like methods on addresses:

```solidity
_destination.functionCall(_payload);
```

For more info on `using A for B` syntax, please read the [documentation](https://docs.soliditylang.org/en/v0.8.21/contracts.html#using-for).

onchain/rollups/contracts/dapp/CartesiDApp.sol
xdaniortega: I will keep that in handles success and failure internally, I'll apply the rest.

onchain/rollups/contracts/dapp/CartesiDApp.sol
xdaniortega: Much cleaner code in this way!

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
```diff
@@ -184,14 +184,14 @@ contract CartesiDAppTest is TestBase {
             LibOutputValidation.getBitMaskPosition(
                 proof.validity.outputIndexWithinInput,
                 _inputIndex
-            )
+            ),
+            abi.encodePacked(uint(1))
```
guidanoli, line 188: Considering `transfer` returns a boolean, it makes more sense to encode `true` rather than `1`, although they have the same ABI-encoding. Also, return data is encoded with `abi.encode` (although in this case, since there are no variable-length values, it won't make a difference).

```suggestion
            abi.encode(true)
```

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
guidanoli, line 1: Also, now that we are propagating errors, we don't need to check that things haven't changed after failed voucher executions because the EVM guarantees that all changes are reverted. I will highlight one example.

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
```diff
@@ -166,8 +166,8 @@ contract CartesiDAppTest is TestBase {
         // not able to execute voucher because dapp has 0 balance
         assertEq(erc20Token.balanceOf(address(dapp)), 0);
         assertEq(erc20Token.balanceOf(recipient), 0);
-        bool success = executeVoucher(voucher, proof);
-        assertEq(success, false);
+        vm.expectRevert("ERC20: transfer amount exceeds balance");
+        executeVoucher(voucher, proof);
         assertEq(erc20Token.balanceOf(address(dapp)), 0);
         assertEq(erc20Token.balanceOf(recipient), 0);
```
guidanoli, line 172: For example, we don't need to check that the balances haven't changed because the call to `executeVoucher` was not successful, which means there were no state changes.
This occurs in many places throughout this file.

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
```diff
@@ -523,14 +521,14 @@ contract CartesiDAppTest is TestBase {
             LibOutputValidation.getBitMaskPosition(
                 proof.validity.outputIndexWithinInput,
                 _inputIndex
-            )
+            ),
+            abi.encodePacked()
```
guidanoli, line 525:
```suggestion
            abi.encode()
```

onchain/rollups/test/foundry/dapp/CartesiDApp.t.sol
```diff
@@ -404,14 +402,14 @@ contract CartesiDAppTest is TestBase {
...
```
guidanoli, line 406:
```suggestion
            abi.encode()
```

guidanoli:
Superseded by #135

PR #109, opened 2023-10-09
feat: decode event data with struct

### LINKED ISSUE #108: Decode event data with structs
## 📚 Context

When testing event emission with forge, we need to decode event data using `abi.decode`. Currently, we are decoding it field-by-field, which is quite cumbersome. The test code it not very readable and prone to errors.

```solidity
(a, b, c) = abi.decode(entry.data, (address, bytes32, address));
```

## ✔️ Solution

In #106, @xdaniortega defined event data as a struct and decoded it field-by-field. This method lets use check each field by their field name.

```solidity
(event.validators, event.quorum) = abi.decode(entry.data, (address[], Quorum));
```

An improvement is to let Solidity do this field-by-field decoding behind the scenes.

```solidity
event = abi.decode(entry.data, (EventData));
```

This pattern can be applied whenever we have a call to `vm.getRecordedLogs()` followed by a loop that check each event.

guidanoli:
Thanks for the insight, @xdaniortega !

Closes #108

### DIFF
--- a/onchain/rollups/test/foundry/consensus/authority/AuthorityHistoryPairFactory.t.sol
+++ b/onchain/rollups/test/foundry/consensus/authority/AuthorityHistoryPairFactory.t.sol
@@ -25,9 +25,24 @@ contract AuthorityFactoryTest is Test {
         IHistoryFactory historyFactory
     );
 
+    struct FactoryCreatedEventData {
+        IAuthorityFactory authorityFactory;
+        IHistoryFactory historyFactory;
+    }
+
     event AuthorityCreated(address authorityOwner, Authority authority);
 
+    struct AuthorityCreatedEventData {
+        address authorityOwner;
+        Authority authority;
+    }
+
     event HistoryCreated(address historyOwner, History history);
+
+    struct HistoryCreatedEventData {
+        address historyOwner;
+        History history;
+    }
 
     event NewHistory(IHistory history);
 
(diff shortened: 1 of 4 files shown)


onchain/rollups/test/foundry/consensus/authority/AuthorityHistoryPairFactory.t.sol
```diff
@@ -148,11 +170,11 @@ contract AuthorityFactoryTest is Test {
             ) {
                 ++numOfNewHistory;
 
-                address a;
+                NewHistoryEventData memory eventData;
 
-                a = abi.decode(entry.data, (address));
+                eventData = abi.decode(entry.data, (NewHistoryEventData));
```
ZzzzHui, line 175: These 2 versions are essentially the same, right? I think we don't have to wrap a struct for every `abi.decode`. Wrapping a struct makes a lot of sense when there are multiple (different) data types. But when there's only 1 (or maybe 2) data type, I don't think we should wrap it. 
Regarding readability, this line could become:
`address history =  abi.decode(entry.data, (address));`

guidanoli: Yes, both versions are isomorphic. We could decode the data into address directly, but if we are using this struct abstraction either way, we might as well be consistent with it, regardless of the number of fields.

xdaniortega: I'll keep it simple, I see the importance of being consistent but it's no sense to use a struct for a simple address decoding.

guidanoli: Fair enough. :-)

ZzzzHui (approved):

PR #116, opened 2023-10-31
feat: improve quorum

guidanoli:
Renames:

- `validatorIndex` :arrow_right: `validatorId`

Adds:

- `validatorById`
- `validators`
- `numOfValidatorsInFavorOf`
- `isValidatorInFavorOf`
- `validatorsInFavorOf`

### DIFF
--- a/onchain/rollups/contracts/consensus/quorum/Quorum.sol
+++ b/onchain/rollups/contracts/consensus/quorum/Quorum.sol
@@ -25,11 +25,11 @@ contract Quorum is AbstractConsensus, PaymentSplitter {
     /// @notice Get the total number of validators.
     uint256 public immutable numOfValidators;
 
-    /// @notice Get the ID of a validator from its address.
+    /// @notice Get the ID of a validator by its address.
     /// @dev Only validators have non-zero IDs.
     mapping(address => uint256) public validatorId;
 
-    /// @notice Get the ID of a validator from its address.
+    /// @notice Get the address of a validator by its ID.
     /// @dev Validator IDs span from 1 to the total number of validators.
     ///      Invalid IDs are assigned to the zero address.
     mapping(uint256 => address) public validatorById;
@@ -127,7 +127,7 @@ contract Quorum is AbstractConsensus, PaymentSplitter {
     /// @notice Check whether a validator is in favor of a claim.
     /// @param _validatorId The ID of the validator
     /// @param _claimData Data for submitting a claim
-    /// @return Array of addresses of validators in favor of claim
+    /// @return Whether validator is in favor of claim
     /// @dev Assumes the provided ID is valid
     function isValidatorInFavorOf(
         uint256 _validatorId,
--- a/onchain/rollups/contracts/consensus/quorum/Quorum.sol
+++ b/onchain/rollups/contracts/consensus/quorum/Quorum.sol
@@ -13,33 +13,41 @@ import {IHistory} from "../../history/IHistory.sol";
 /// @title Quorum consensus
 /// @notice A consensus model controlled by a small set of addresses, the validators.
 ///         In this version, the validator set is immutable.
-/// @dev This contract uses OpenZeppelin `PaymentSplitter` and `BitMaps`.
+///         Claims are stored in an auxiliary contract called history.
+/// @dev Each validator is assigned an identifier that spans from 1 to N,
+///      where N is the total number of validators in the quorum.
+///      These identifiers are used internally instead of addresses for optimization reasons.
+///      This contract uses OpenZeppelin `PaymentSplitter` and `BitMaps`.
 ///      For more information on those, please consult OpenZeppelin's official documentation.
 contract Quorum is AbstractConsensus, PaymentSplitter {
     using BitMaps for BitMaps.BitMap;
 
+    /// @notice Get the total number of validators.
+    uint256 public immutable numOfValidators;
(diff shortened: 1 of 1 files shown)


onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -69,25 +82,99 @@ contract Quorum is AbstractConsensus, PaymentSplitter {
...
-            if (++claimVotes.count == 1 + numOfValidators / 2) {
+            if (++votingStatus.inFavorCount == 1 + numOfValidators / 2) {
                 history.submitClaim(_claimData);
             }
         }
     }
 
+    /// @notice Get an array with the IDs of all validators.
+    /// @return Array of IDs of validators
+    function validatorIds() external view returns (uint256[] memory) {
```
ZzzzHui, line 105: What's the use case for getting an array with ids? Is it sufficient to get `numOfValidators` instead?

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -69,25 +82,99 @@ contract Quorum is AbstractConsensus, PaymentSplitter {
...
+        bytes calldata _claimData
+    ) external view returns (uint256) {
+        VotingStatus storage votingStatus = votingStatuses[_claimData];
+        return votingStatus.inFavorCount;
+    }
+
+    /// @notice Check whether a validator is in favor of a claim.
+    /// @param _validatorId The ID of the validator
+    /// @param _claimData Data for submitting a claim
+    /// @return Array of addresses of validators in favor of claim
```
ZzzzHui, line 144: This return statement is a duplicate

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -69,25 +82,99 @@ contract Quorum is AbstractConsensus, PaymentSplitter {
...
+    /// @dev Assumes the provided ID is valid
+    function isValidatorInFavorOf(
+        uint256 _validatorId,
+        bytes calldata _claimData
+    ) external view returns (bool) {
+        VotingStatus storage votingStatus = votingStatuses[_claimData];
+        BitMaps.BitMap storage inFavorById = votingStatus.inFavorById;
```
ZzzzHui, line 151: Should we make these 2 lines into 1?
```
BitMaps.BitMap storage inFavorById = votingStatuses[_claimData].inFavorById;
```

ZzzzHui (commented):
Now that we have a new variable `validatorById`, the original variable `validatorId` seems redundant to me, because a validator's id could simply be its key in `validatorById`. 

Plus, the id doesn't need to start from 1 now. We started the id from 1 because we only used  `mapping(address=>uint256)` and want to differentiate an address being a validator and not. But now if we have `validatorById` then only address being a validator (except address 0) will be in this mapping. So we can start id from 0 now.

If we remove variable `validatorId`, when calling function `submitClaim`, we would need to provide `_validatorId`, like the function `isValidatorInFavorOf` does. And then check if `msg.sender == validatorById[_validatorId]`.

What do you think

onchain/rollups/contracts/consensus/quorum/Quorum.sol
```diff
@@ -13,33 +13,41 @@ import {IHistory} from "../../history/IHistory.sol";
...
-    struct Votes {
-        // how many has voted
-        uint256 count;
-        // use BitMap to record who has voted
-        BitMaps.BitMap votedBitMap;
+    /// @notice Get the ID of a validator from its address.
+    /// @dev Only validators have non-zero IDs.
+    mapping(address => uint256) public validatorId;
+
+    /// @notice Get the ID of a validator from its address.
```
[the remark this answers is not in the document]
guidanoli: Oops, you're right!

onchain/rollups/contracts/consensus/quorum/Quorum.sol
guidanoli: Yes, we say in the documentation that IDs range from 1 to `numOfValidators`.
Maybe we should remove this function. :-)

onchain/rollups/contracts/consensus/quorum/Quorum.sol
guidanoli: True, sorry!

onchain/rollups/contracts/consensus/quorum/Quorum.sol
guidanoli: We could, but then `prettier` would break the line in a not-so-pretty way.

ZzzzHui: I see. Do you think it would be more readable even though broken by prettier? Or it doesn't make that much of a difference

ZzzzHui:
I agree that a single $O(n)$ cost during deployment is likely worthy than providing its own id every time to submit claim, and knowing its id by one time search or listening to event etc. Yep, so let's keep the variable  `validatorId`

onchain/rollups/contracts/consensus/quorum/Quorum.sol
guidanoli: I think it's more readable the way it currently is now, broken in two lines. :-)

ZzzzHui (approved):

PR #119, opened 2023-11-02
Remove yarn workspace

guidanoli:
We want to use `changesets` for managing releases, but if `@cartesi/rollups` is contained inside a yarn workspace, then the tag that `changesets` would create would be like `@cartesi/rollups@1.2.0`, while we want it to be `v1.2.0`. In order to do that, we need to make `@cartesi/rollups` no longer part of a yarn workspace, but a package of its own. So, this is the first change.

**Remove yarn workspace:**
- Update `check-code-format.yml` to run `yarn` inside `onchain/rollups`
- Move `.changeset` to `onchain/rollups`
- Move lines from `onchain/.gitignore` to `onchain/rollups/.gitignore`
- Fix references to `../node_modules` inside `onchain/rollups`

Taking advantage of this massive change, I also took the freedom to...
- Bump `yarn` to its latest version
- Bump all actions to their latest versions
- Remove the `cog.toml` file (and create it only in the CI)
- Rename `build.yml` to `test.yml`

A second change was the removal of `publish.yml` and addition of `release.yml`.
While `publish.yml` runs `yarn npm` directly, `release.yml` uses `changesets/action`.
This command does two things:
- Creates a PR that removes all changeset files and updates `package.json` and `CHANGELOG.md`
- Publishes releases to npm (`@cartesi/rollups`) and to GitHub (tag and release)

Finally, I needed to move the `CHANGELOG.md` from the repository root to `onchain/rollups`, because the `changesets` action expects this file to be in the package root directory. Later, when `rollups-arbitration` gets merged into `rollups`, we'll move everything inside `rollups` to the repository root, so `CHANGELOG.md` will come back to the repository root. Also, I filtered the entries from `CHANGELOG.md` that were not related to `@cartesi/rollups` (but were related to the node, for example), and formatted it like `changesets` does it.


README.md
```diff
@@ -34,29 +34,27 @@ This repository uses [Yarn](https://yarnpkg.com/getting-started/install) to mana
...
+cd onchain/rollups
 yarn install
 ```
 
 ## Tests
 
 If you plan to run the [Forge](https://book.getfoundry.sh/getting-started/installation) tests, there still some setup left to do.
-Assuming you are on the `onchain` directory, and that [Docker Engine](https://docs.docker.com/get-docker/) is running on the background, you may run the following commands.
+Assuming you are on the `onchain/rollups` directory, and that [Docker Engine](https://docs.docker.com/get-docker/) is running on the background, you may run the following command.
+This command will build the Cartesi Machine image necessary to build the proofs.
```
ZzzzHui, line 45: should we say "Cartesi Virtual Machine"?

guidanoli: This is actually a great topic of discussion. I personally prefer "Cartesi Machine".
https://discord.com/channels/600597137524391947/1169240346564251750

onchain/rollups/.changeset/blue-schools-think.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
ZzzzHui, line 2: this changeset file has nothing in it?

guidanoli: Yes, this is necessary because otherwise, when I run...

```
$ yarn changeset status
🦋  error Some packages have been changed but no changesets were found. Run `changeset add` to resolve this error.
🦋  error If this change doesn't need a release, run `changeset add --empty`.
```

PR #120, opened 2023-11-03
Fixing the readability of Partition.sol

Aryan9592:
After opening the file in github, at first it seems ok but as we scroll down the code..it suddenly starts appearing of blue color which really seems odd. Atleast that's what I faced while going through the code. Just fixed it by removing the ' on line 81

### DIFF
--- a/onchain/rollups-arbitration/src/partition/Partition.sol
+++ b/onchain/rollups-arbitration/src/partition/Partition.sol
@@ -78,7 +78,7 @@ library Partition {
         WaitingHash memory waitingHash =
             advancePartition(waitingInterval, agree);
 
-        require(mustEndPartition(waitingHash), "can't end partition yet");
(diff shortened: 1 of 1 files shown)


guidanoli:
I suspect this may have something to do with the syntax highlight software you are using.

Aryan9592:
Yeah, I found this lot in other repositories too. and Many have gone with this fix.
Don't know when GitHub gonna fix this one!

guidanoli:
Since this is an issue with the GitHub front-end and not with the contract itself, may we close this PR?

Aryan9592:
Well, you can. But don't you think readability matters till Github fixes this issue?
It's like we really don't want to see the **blue lines** just cuz Github having a issue with this. Why not move with something which fixes this one temporarily rather than relying on Github. Later, you can revert that `cant` to `can't` if that is really important for this contract. 
Just my view, rest depends on you...Thanks @guidanoli

PR #106, opened 2023-10-09
Introducing Quorum Factory

xdaniortega:

PR #121, opened 2023-11-03
feat: add quorum factory and factory tests

ZzzzHui:


ZzzzHui:
Ohhhh. Totally forgot that Daniel was working on this issue and had a branch created [here](https://github.com/cartesi/rollups-contracts/pull/106/files#diff-e1ce3ebf0f6fb15d4f4c9d20130e8a1712df663f68347f4be848a93f91058a2e). I went ahead and created my own branch. The base of this branch is more recent so we might focus on this branch then.

onchain/.changeset/stupid-doors-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cartesi/rollups": minor
+---
+
+add quorum factory and factory tests
```
guidanoli, line 5: I think we should seek consistency in the changelog.
So, let me propose some major guidelines:

- Capitalize the first word of sentences
- Format code-like text (like `this`)
- Conjugate action verbs in the past tense

Also, tests won't make it to the `@cartesi/rollups` package, so they can be abstracted away from the changelog. We might, however, mention tests when they affect the actual contracts, for example, when they expose a new bug.

```suggestion
Added `QuorumFactory` contract and `IQuorumFactory` interface
```

onchain/rollups/contracts/consensus/quorum/IQuorumFactory.sol
```diff
@@ -0,0 +1,61 @@
...
+
+/// @title Quorum Factory interface
+interface IQuorumFactory {
+    // Events
+
+    /// @notice A new quorum was deployed.
+    /// @param quorum The quorum
+    /// @param history The history
+    /// @dev MUST be triggered on a successful call to `newQuorum`.
+    event QuorumCreated(Quorum quorum, IHistory history);
```
guidanoli, line 17: Why not emit `validators` and `shares` as well?

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
```diff
@@ -0,0 +1,121 @@
...
+    uint256[] shares;
+    uint256 numValidator;
+
+    event QuorumCreated(Quorum quorum, IHistory history);
+
+    function setUp() public {
+        factory = new QuorumFactory();
+    }
+
+    function testNewQuorum(uint16 _numValidator, IHistory _history) public {
```
guidanoli, line 25: We can have `_numValidator` as a `uint256` fuzzy parameter, and use [`bound`](https://book.getfoundry.sh/reference/forge-std/bound) from `forge-std` to restrict its domain to the desired $(0,1000)$ range. This should also improve performance of this fuzzy test case, as the function will go from a $998/2^{16} \approx 1.5\\%$ chance of passing through the `vm.assume` check to $100\\%$ (no `vm.assume`).

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
```diff
@@ -0,0 +1,121 @@
...
+import {QuorumFactory} from "contracts/consensus/quorum/QuorumFactory.sol";
+import {Quorum} from "contracts/consensus/quorum/Quorum.sol";
+import {IHistory} from "contracts/history/IHistory.sol";
+import {Vm} from "forge-std/Vm.sol";
+
+contract QuorumFactoryTest is Test {
+    QuorumFactory factory;
+    address[] validators;
+    uint256[] shares;
+    uint256 numValidator;
```
guidanoli, line 17: This state variable can be removed, and any usage of it can be substituted by the length of either array.

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
```diff
@@ -0,0 +1,121 @@
...
+
+    function generateAddrAndShares(uint256 _numValidator) internal {
+        // assume there are less than 1000 validators so the test wont run forever
+        vm.assume(_numValidator > 0 && _numValidator < 1000);
+        numValidator = _numValidator;
+
+        validators = new address[](numValidator);
+        shares = new uint256[](numValidator);
+        for (uint256 i; i < numValidator; ++i) {
+            validators[i] = vm.addr(i + 1); // assume there is no address collision
```
guidanoli, line 43: We can guarantee that there will be no collision if addresses are assigned linearly. We know this will not create collisions because $i + 1 \le numValidator < 1000 < 2^{160}$.

```suggestion
            validators[i] = address(uint160(i + 1));
```

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
```diff
@@ -0,0 +1,121 @@
...
+            shares[i] = i + 1;
```
guidanoli, line 44: Maybe we could hash `i`, cast it into a `uint256`, and force it to be contained in the $(1,(2^{256}-1)/n)$ interval with `bound`, where $n$ is the number of validators. This way, we guarantee that...

$$
(\forall i \in [1,n], shares_i > 0) \wedge \sum_{i=1}^n shares_i < 2^{256}
$$

... which is required by `PaymentSplitter` [(1)](https://github.com/binodnp/openzeppelin-solidity/blob/ffc4bcbba398b61d19ede03e4fdf673ddbce1083/contracts/payment/PaymentSplitter.sol#L108) [(2)](https://github.com/binodnp/openzeppelin-solidity/blob/ffc4bcbba398b61d19ede03e4fdf673ddbce1083/contracts/payment/PaymentSplitter.sol#L113).

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
```diff
@@ -0,0 +1,121 @@
...
+                entry.emitter == address(factory) &&
+                entry.topics[0] == QuorumCreated.selector
+            ) {
+                ++numQuorumCreated;
+                (address quorumAddr, address historyAddr) = abi.decode(
+                    entry.data,
+                    (address, address)
+                );
+                assertEq(quorumAddr, address(_quorum));
+                assertEq(historyAddr, address(_history));
```
guidanoli, line 70: We could adopt #108 here.

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
```diff
@@ -0,0 +1,121 @@
...
+import {Test} from "forge-std/Test.sol";
+import {QuorumFactory} from "contracts/consensus/quorum/QuorumFactory.sol";
+import {Quorum} from "contracts/consensus/quorum/Quorum.sol";
+import {IHistory} from "contracts/history/IHistory.sol";
+import {Vm} from "forge-std/Vm.sol";
+
+contract QuorumFactoryTest is Test {
+    QuorumFactory factory;
+    address[] validators;
+    uint256[] shares;
```
guidanoli, line 16: These state variables can be removed. Instead of storage, they can be (and kind of already are, really) handled in memory! For example, `generateAddrAndShares` should return these, instead of writing them to storage. This should allow this function to be `pure` then.

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
```diff
@@ -0,0 +1,121 @@
...
+                );
+                assertEq(quorumAddr, address(_quorum));
+                assertEq(historyAddr, address(_history));
+            }
+        }
+        assertEq(numQuorumCreated, 1);
+
+        assertEq(_quorum.numOfValidators(), numValidator);
+        assertEq(address(_quorum.getHistory()), address(_history));
+        address[] memory returnedValidators = _quorum.validators();
```
guidanoli, line 77: We should check here that:
- the array of validators returned by the Quorum contract (via `validators`) should have the expected length (`numValidator`)
- the total number of shares returned by the Quorum contract (via `totalShares()`) should be the expected value (sum of `_shares`)

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
guidanoli, line 1: To stay consistent with the Quorum contract interface, we should rename `numValidator` as `numOfValidators`.

guidanoli:
This will need a rebase.
Do note that the `.changeset` directory was moved, changeset file should be moved as well. :-)

onchain/rollups/contracts/consensus/quorum/IQuorumFactory.sol
ZzzzHui: Because they were already emitted in the Quorum contract via PaymentSplitter, right? 
ref: https://github.com/binodnp/openzeppelin-solidity/blob/ffc4bcbba398b61d19ede03e4fdf673ddbce1083/contracts/payment/PaymentSplitter.sol#L114C35-L114C35

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
ZzzzHui: Yes we can. Although it's more concise to say `numValidator` compared with `_validators.length`. So even after removing this state variable, I would probably still create a local variable whichever test it appears in. And since it's a test contract, we don't need to worry about gas optimization. So one state variable is easier than potentially multiple local variables? What do you think?

onchain/rollups/contracts/consensus/quorum/IQuorumFactory.sol
guidanoli: Okay, fair enough.

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
guidanoli: I think we should only store the array in storage and access its length locally in each function, this way we avoid the two variables going out-of-sync.

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
ZzzzHui: To sum an arithmetic sequence to $2^{256}$, we would need to have about $2^{128}$ validators. I don't think the quorum would get that large

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
ZzzzHui: We could do that, although gas optimization is not an issue in test contract.

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
guidanoli: > To sum an arithmetic sequence to 2256, we would need to have about 2128 validators. I don't think the quorum would get that large

Yes, but if you hash each element of the arithmetic sequence, you might get some large numbers that can add up to more than $2^{256}$. Either way, I might be overcomplicating things and using the arithmetic sequence like you are already doing.

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
guidanoli: It's not much about gas, but about where data is stored.
I think the code would look much cleaner if it were handled in memory.

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
```diff
@@ -0,0 +1,137 @@
...
+        assertEq(returnedValidators.length, _validators.length);
+        uint256 totalShares;
+        for (uint256 i; i < _validators.length; ++i) {
+            assertEq(returnedValidators[i], _validators[i]);
+            assertEq(_quorum.shares(returnedValidators[i]), _shares[i]);
+            totalShares += _shares[i];
+        }
+        assertEq(_quorum.totalShares(), totalShares);
```
guidanoli, line 94: I've pushed a change to `feature/quorum` which lets you calculate the sum of arrays more easily. This can simplify your code.

https://github.com/cartesi/rollups-contracts/blob/a8454c6ecd38f2cb7a84def3625bd5d369579f18/onchain/rollups/test/foundry/consensus/quorum/Quorum.t.sol#L122

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
```diff
@@ -0,0 +1,137 @@
...
+        view
+        returns (address[] memory validators_, uint256[] memory shares_)
+    {
+        // let there be no more than 1000 validators so the test wont run forever
+        uint256 numOfValidators = bound(_seed, 1, 1000);
+
+        validators_ = new address[](numOfValidators);
+        shares_ = new uint256[](numOfValidators);
+        for (uint256 i; i < numOfValidators; ++i) {
+            validators_[i] = address(uint160(i + 1));
```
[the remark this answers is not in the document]
ZzzzHui: Yeah. Either way is fine. There's no difference really :)

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
```diff
@@ -0,0 +1,137 @@
...
+
+        testNewQuorumAux(validators, shares, _history, quorum);
+    }
+
+    function generateAddrAndShares(
+        uint256 _seed
+    )
+        internal
+        view
+        returns (address[] memory validators_, uint256[] memory shares_)
```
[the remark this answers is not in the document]
ZzzzHui: I think it's simpler to be 1 function. There's no demand to generate solely the array of validators, or solely the array of shares. If we split it into 2 functions, the main function bodies of them would be the same, thus increasing code redundancy. 
As for limiting to `uint8`. I did intentionally choose a number greater than 255 in the test, because BitMap uses a new storage slot for a quorum greater than 255. Although in this factory tests, it doesn't have any effect because we don't submit claims or anything. Just to explain why I had in mind a quorum greater than 255. I think it's more flexity to not limit the quorum size to `uint8.max`

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
```diff
@@ -0,0 +1,137 @@
...
+        }
+        assertEq(numQuorumCreated, 1);
+
+        assertEq(_quorum.numOfValidators(), _validators.length);
+        assertEq(address(_quorum.getHistory()), address(_history));
+        address[] memory returnedValidators = _quorum.validators();
+        assertEq(returnedValidators.length, _validators.length);
+        uint256 totalShares;
+        for (uint256 i; i < _validators.length; ++i) {
+            assertEq(returnedValidators[i], _validators[i]);
```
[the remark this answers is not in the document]
ZzzzHui: Nice catch 👍

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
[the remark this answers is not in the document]
guidanoli: Oh, good catch. I think I will update the quorum tests to encompass that then!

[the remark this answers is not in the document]
ZzzzHui: But then we would have 2 function bodies that have basically the same code logic, which adds redundancy.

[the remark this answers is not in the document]
ZzzzHui: Unless we want to have those individual functions in the `util/TestBase.sol`. Then it makes sense to modularize them. What do you think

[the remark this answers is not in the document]
guidanoli: Yes, let's move them to `TestBase`:
- `generateAddresses(uint256 n) -> address[]`
- `generateArithmeticSequence(uint256 a0, uint256 d, uint256 n) -> uint256[]`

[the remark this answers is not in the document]
guidanoli: In fact, let me do that on `feature/quorum`. I can use these for the Quorum tests as well!

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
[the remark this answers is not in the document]
guidanoli: Now, we can remove this function and use `generateAddresses` and `generateArithmeticSequence` instead. :-)

onchain/rollups/contracts/consensus/quorum/QuorumFactory.sol
```diff
@@ -0,0 +1,57 @@
...
+
+        emit QuorumCreated(quorum, _history);
+
+        return quorum;
+    }
+
+    function calculateQuorumAddress(
+        address[] calldata _validators,
+        uint256[] calldata _shares,
+        IHistory _history,
```
guidanoli, line 43: We are not using `IHistory` interface for anything in this function body, and we could just as well type this argument as `address`. By doing so, the `QuorumHistoryPairFactory` won't need to cast an address to `IHistory`, which I find worse.

onchain/rollups/test/foundry/consensus/quorum/QuorumFactory.t.sol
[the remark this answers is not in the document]
ZzzzHui: Yep. I thought the same too :)

guidanoli:
This needs a rebase.

guidanoli (approved):
LGTM

PR #129, opened 2023-11-08
Update MemoryManager.sol

Aryan9592:
Here, we can upgrade this for loop and make it even more efficient by introducing these changes.

Refer this [gist](https://gist.github.com/grGred/9bab8b9bad0cd42fc23d4e31e7347144#for-loops-improvement) and [PR](https://github.com/nomoixyz/vulcan/pull/177)


guidanoli:
We'll soon be bumping `solc` to 0.8.23, which will include a compiler feature that removes the need for unchecked for loop counter increments. Thanks for the PR, still! :-)

PR #135, opened 2023-11-09
feat: propagate errors on voucher execution

### LINKED ISSUE #96: Propagate errors raised by low-level calls
## 📚 Context

In order to allow DApp developers to perform any conceivable message call from the context of a DApp, we allow vouchers to execute a low-level message call. The return value of such a low-level call are a boolean flag (whether the call was successful or reverted) and some data (which can be either the return values of a function call or error data, depending on the value of the aforementioned success flag).

Currently, we are only checking whether the call was successful and ignoring the returned data altogether. This means that any error data which could be valuable for DApp developers to debug is suppressed and doesn't make it to the UI.

## ✔️ Solution

One solution is to propagate the error to the user. We could do this ourselves, but I've just recently found that OpenZeppelin implements this in their [`Address.sol`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Address.sol) library. The nice thing about this library is that it also implements this error handling for `DELEGATECALL`, which we hope to use with the [Output Unification](#42).

As a result, any call to `executeVoucher` with enough gas either returns `true` or propagates the error originally raised by the sub-call. Smart contracts can still do error handling by means of try-and-catch clauses.

guidanoli:
Fixes #96

### DIFF
--- a/onchain/rollups/contracts/dapp/CartesiDApp.sol
+++ b/onchain/rollups/contracts/dapp/CartesiDApp.sol
@@ -12,6 +12,7 @@ import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
 import {ERC721Holder} from "@openzeppelin/contracts/token/ERC721/utils/ERC721Holder.sol";
 import {ERC1155Holder} from "@openzeppelin/contracts/token/ERC1155/utils/ERC1155Holder.sol";
 import {ReentrancyGuard} from "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
 
 /// @title Cartesi DApp
 ///
@@ -64,6 +65,7 @@ contract CartesiDApp is
 {
     using Bitmask for mapping(uint256 => uint256);
     using LibOutputValidation for OutputValidityProof;
+    using Address for address;
 
     /// @notice Raised when executing an already executed voucher.
     error VoucherReexecutionNotAllowed();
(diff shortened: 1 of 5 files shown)


ZzzzHui (dismissed):

onchain/rollups/contracts/dapp/CartesiDApp.sol
```diff
@@ -134,15 +136,11 @@ contract CartesiDApp is
         }
 
         // execute voucher
```
pedroargento, line 138: Should we add an comment here annotating that it will revert if the `functionCall` fails? As the syntax does not make clear that its an Open Zeppelin import and the behavior is different from the usual call.

guidanoli: This can be easily deducted from the fact that `functionCall` is not a built-in [member of addresses](https://docs.soliditylang.org/en/latest/types.html#members-of-addresses), and that we are using the `Address` library from OpenZeppelin for `address` values. [(1)](https://github.com/cartesi/rollups-contracts/blob/515477f6fff67c308153e5581388910cb84d40cf/onchain/rollups/contracts/dapp/CartesiDApp.sol#L15) [(2)](https://github.com/cartesi/rollups-contracts/blob/515477f6fff67c308153e5581388910cb84d40cf/onchain/rollups/contracts/dapp/CartesiDApp.sol#L68)

Also, I think it is unnecessary to say that a call to another contract might revert, and that any error will be propagated, beacuse this is the default behavior of calls in Solidity. Errors must be handled explicitly, either by low-level calls or through try-catch clauses.

pedroargento (approved):

ZzzzHui (approved):

PR #136, opened 2023-11-10
feat: reject failed ERC-20 transfers

### LINKED ISSUE #113: Make ERC-20 portals reject failed transfers again
## 📚 Context

The job of the ERC-20 portal is to manage ERC-20 token deposits to the DApp. It does so by calling the `transferFrom` function on the token contract and adding an input in the DApp's input box. This function returns a boolean value which signals if the transfer was successful or not.

Prior to 0.9.0, the ERC-20 portal required this boolean value to be `true`, which means that the DApp would only be notified of an ERC-20 if it were successful.

On 0.9.0, we decided to accept transfers in which the `transferFrom` function returned false, and added this information in the input itself. The rationale was that EIP-20 did not specify clearly the semantics of this value, and we would like the DApp back-end to handle this information however they want.

On October 25th, however, @gligneul pointed out that this behavior could open the possibility for DApp developers to mistakenly accept failed transfers, should they not check this boolean value within the input. @ZzzzHui, @diegonehab and I agreed.

## ✔️ Solution

Make the `depositERC20Tokens` function revert if `transferFrom` returns `false`.
We may keep the input structure for backwards compatibility.

guidanoli:
Closes #113

### DIFF
--- a/onchain/rollups/.changeset/plenty-feet-help.md
+++ b/onchain/rollups/.changeset/plenty-feet-help.md
@@ -1,0 +1,7 @@
+---
+"@cartesi/rollups": major
+---
+
+Removed the boolean `success` flag from ERC-20 deposit inputs.
+This change was made because this flag would always be `true`, giving no extra information to the DApp back-end.
+Consumers of this input should be aware of the new encoding.
--- a/onchain/rollups/.changeset/tough-shoes-taste.md
+++ b/onchain/rollups/.changeset/tough-shoes-taste.md
@@ -1,0 +1,7 @@
+---
+"@cartesi/rollups": major
+---
+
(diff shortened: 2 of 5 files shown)


pedroargento (dismissed):

ZzzzHui (dismissed):

guidanoli:
I am going to change the implementation to use [`SafeERC20`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.9.2/contracts/token/ERC20/utils/SafeERC20.sol) from OpenZeppelin.

ZzzzHui (approved):

pedroargento (approved):

ZzzzHui (approved):

pedroargento:
LGTM!

PR #163, opened 2023-12-01
refactor!: remove `withdrawERC20Tokens` function

### LINKED ISSUE #162: Remove ERC-20 entry points from consensus contract(s)
## 📚 Context

Currently, the `Authority` contract defines a function called `withdrawERC20Tokens` which is meant to serve as an entry point for the owner to withdraw ERC-20 tokens from the `Authority` contract to its own balance. This feature was added with the idea that the `Authority` contract would be the recipient of ERC-20 tokens as payment for validating DApps. In reality, however, this function is not being used by anyone, and has, instead, created more problems than solved.

One issue is that it unnecessarily complicates the consensus contract with a financial layer which could live elsewhere. For example, the `Authority` owner could ask for ERC-20 tokens to be deposited directly in the owner's account, instead of the contract's.

Another issue is that, for the `Quorum` consensus, we would use OpenZeppelin's `PaymentSplitter` contract to split the payment of ERC-20 tokens amongst quorum members according to some distribution of shares. However, we'd like to upgrade `@openzeppelin/contracts` to v5 (see #161), which has removed such contract, leaving no equivalent alternative.

## ✔️ Solution

We should remove the `withdrawERC20Tokens` function from the `Authority` contract.
As for `Quorum`, we don't need to implement any type of financial layer either.
The financial layer can live on top of currently existing contracts.

guidanoli:
Closes #162


onchain/rollups/.changeset/short-bears-flash.md
```diff
@@ -0,0 +1,8 @@
+---
+"@cartesi/rollups": major
+---
+
+Removed the `withdrawERC20Tokens` function from `Authority` contract.
+This function was removed due to the lack of usage, and because implementing a similar function for `Quorum` would not be possible with OpenZeppelin v5.
```
ZzzzHui, line 6: Maybe we should say 5.0

guidanoli: Good point.

pedroargento (approved):

ZzzzHui (approved):
Should there be 2 changeset files for 1 commit?

guidanoli:
> Should there be 2 changeset files for 1 commit?

It's because these are different changes. One is about the function and the other is about the error. I could have put them into the same changeset file too.

PR #176, opened 2023-12-15
feat!: change `VoucherExecuted` event

### LINKED ISSUE #169: Use OpenZeppelin's `BitMaps` in `CartesiDApp`
## 📚 Context

Currently, we are using a homebrew implementation of bit masks called [`Bitmask`](https://github.com/cartesi/solidity-util/blob/v6.0.0/packages/contracts/contracts/Bitmask.sol) for keeping track of vouchers that were already executed. However, there is an equivalent OpenZeppelin implementation called [`BitMaps`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.0.0/contracts/utils/structs/BitMaps.sol).

## ✔️ Solution

Refactor `CartesiDApp` to use `BitMaps` instead of `Bitmask`.

### LINKED ISSUE #170: Refactor `VoucherExecuted` event
## 📚 Context

Currently, the `VoucherExecuted` event from the `ICartesiDApp` interface has one argument, `voucherId`, which uniquely identifies the voucher that was executed.
This number, however, isn't really helpful for off-chain components to tell which voucher was actually executed.

## ✔️ Solution

Vouchers can be identified with two numbers: input index and output index (within input).
We can refactor the `VoucherExecuted` event to encode the two numbers mentioned as arguments.

### LINKED ISSUE #171: Simplify and optimize voucher executed flag storage
## 📚 Context

Currently, we are storing whether a voucher was executed or not in a bit mask.
The 256-bit key to this bit mask is a concatenation of the 128-bit input index and 128-bit output index (within input).
This method is not only potentially unsafe but unnecessarily complex.

## ✔️ Solution

We can maintain the locality of bits among vouchers of the same output index, while simplifying on-chain.
So, instead of having a single bit mask for all vouchers, we can have a mapping of bit masks, index by the output index.
We index first by output index because inputs normally generate only a few vouchers.
With this, there will not be the need to calculate voucher "positions" in the bit mask.

guidanoli:
Closes #169
Closes #170
Closes #171


pedroargento (approved):

ZzzzHui (approved):

guidanoli:
FYI @tuler We won't be using `Bitmask` for v2 anymore.

PR #199, opened 2024-01-12
update readme to include quorum

ZzzzHui:


README.md
```diff
@@ -220,7 +220,9 @@ Notices are informational statements that can be proved by contracts in the base
 
 This module is responsible for providing valid claims to DApps after reaching some form of consensus. Each DApp has its own mapping of claims, each of which is mapped by the range of input indices of an epoch.
 
-The module's interface aims to be as generic as possible to accommodate any consensus model, since there are plenty to choose from. One type of consensus implemented by Cartesi is called Authority. It is owned by a single address, who has complete power over the consensus. It is arguably the simplest consensus to implement, although quite vulnerable.
+The module's interface aims to be as generic as possible to accommodate any consensus model, since there are plenty to choose from. The types of consensus implemented by Cartesi include:
```
guidanoli, line 223: I think it's good to point out that we might implement more in the future, by adding the word "currently". Also, I think it may sound weird to say "by Cartesi" given the current descentralization roadmap.

```suggestion
The module's interface aims to be as generic as possible to accommodate any consensus model, since there are plenty to choose from. The types of consensus currently implemented include:
```

README.md
```diff
@@ -220,7 +220,9 @@ Notices are informational statements that can be proved by contracts in the base
...
+- Authority: owned by a single address, who has complete power over the consensus. It is arguably the simplest consensus to implement, although quite vulnerable.
+- Quorum: comprised of a set of validators. Consensus is achieved if the majority of validators agreed on the same claim.
```
guidanoli, line 225: Let's use the verb "manage" for both consensuses for consistency.
I would change "the simplest" to just "trivial", because an even simpler consensus to implement would be the permissionless consensus, where anyone can claim anything.
Also, I think the word "consensus" goes better with the verb "reach" rather than with the verb "achieve".

```suggestion
- Authority: managed by a single address, who has complete power over the consensus. It is trivial to implement, yet quite vulnerable.
- Quorum: managed by a generally small, finite set of validators. Consensus is reached when the majority of the quorum agrees on any given claim.
```

pedroargento (approved):

guidanoli (approved):

PR #207, opened 2024-01-21
chore: adjust changesets

guidanoli:


guidanoli:
I'm trying to make the changelog looks less bloated.
Also, I recently found out we can use lists!

onchain/rollups/.changeset/short-bears-flash.md
```diff
@@ -2,7 +2,14 @@
...
+Removed from the `Authority` contract:
+
+-   `AuthorityWithdrawalFailed` error
+-   `NewHistory` event
+-   `getClaim` function
+-   `getHistory` function
+-   `join` function
+-   `migrateHistoryToConsensus` function
+-   `setHistory` function
+-   `submitClaim(bytes)` function
```
ZzzzHui, line 14: this function is changed with different parameters. We can either put a note for that, or as is is good too.

ZzzzHui (approved):

onchain/rollups/.changeset/short-bears-flash.md
guidanoli: Ok, I will add a note.

ZzzzHui (approved):

PR #210, opened 2024-01-22
refactor!: type input indices as `uint64`

### LINKED ISSUE #209: Type input indices as `uint64`
## 📚 Context

The EVM word is 256 bits long. That is why `uint256` is so ubiquitous in Solidity. Outside the EVM, however, 256-bit unsigned integers are not that common to find. In some cases, 256 bits may be a bit too much. UNIX timestamps, for example, will overflow with 32-bit signed integers in 2038, and with 32-bit unsigned integers in 2106. Input indices don't need 256 bits. The Ethereum blockchain, for example, haven't even reached $2^{32}$ yet.

## ✔️ Solution

Type input indices as `uint64`.

## :brain: Rationale

We assume applications won't receive more than $2^{64}$ inputs.
With this, we can save on-chain space and avoid `uint256` in off-chain components.

guidanoli:
Close #209

### DIFF
--- a/onchain/rollups/contracts/dapp/Application.sol
+++ b/onchain/rollups/contracts/dapp/Application.sol
@@ -22,6 +22,7 @@ import {Address} from "@openzeppelin/contracts/utils/Address.sol";
 import {IERC721Receiver} from "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
 import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
 import {BitMaps} from "@openzeppelin/contracts/utils/structs/BitMaps.sol";
+import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 
 contract Application is
     IApplication,
@@ -35,6 +36,7 @@ contract Application is
     using LibProof for Proof;
     using LibInputRange for InputRange;
     using Address for address;
+    using SafeCast for uint256;
 
     /// @notice The initial machine state hash.
     /// @dev See the `getTemplateHash` function.
@@ -123,7 +125,7 @@ contract Application is
         // reverts if proof isn't valid
         proof.validity.validateVoucher(destination, payload, epochHash);
 
-        uint256 outputIndexWithinInput = proof.validity.outputIndexWithinInput;
+        uint64 outputIndexWithinInput = proof.validity.outputIndexWithinInput;
         BitMaps.BitMap storage bitmap = _voucherBitmaps[outputIndexWithinInput];
 
         // check if voucher has been executed
@@ -136,7 +138,7 @@ contract Application is
 
         // mark it as executed and emit event
         bitmap.set(inputIndex);
-        emit VoucherExecuted(inputIndex, outputIndexWithinInput);
+        emit VoucherExecuted(inputIndex.toUint64(), outputIndexWithinInput);
     }
 
     function migrateToConsensus(
--- a/onchain/rollups/contracts/library/LibProof.sol
+++ b/onchain/rollups/contracts/library/LibProof.sol
@@ -10,7 +10,7 @@ library LibProof {
         Proof calldata proof
     ) internal pure returns (uint256 inputIndex) {
         inputIndex =
-            proof.inputRange.firstIndex +
+            uint256(proof.inputRange.firstIndex) +
             proof.validity.inputIndexWithinEpoch;
     }
 }
--- a/onchain/rollups/contracts/common/InputRange.sol
+++ b/onchain/rollups/contracts/common/InputRange.sol
@@ -7,6 +7,6 @@ pragma solidity ^0.8.8;
 /// @param firstIndex The index of the first input
 /// @param lastIndex The index of the last input
 struct InputRange {
-    uint256 firstIndex;
-    uint256 lastIndex;
+    uint64 firstIndex;
+    uint64 lastIndex;
 }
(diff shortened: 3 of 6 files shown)


ZzzzHui:
Even if we want to save space for events, we don't need to use `uint64` in every intermediary step, for example, local variables and function arguments. Because in those intermediary steps, if we cast them as `uint64`, they would be auto cast back to `uint256` for processing. So we should still use `uint256` as much as possible.
Plus, events doesn't cost too much gas, ref: https://github.com/wolflo/evm-opcodes/blob/main/gas.md#a8-log-operations So if we run gas report on this PR, we can see gas costs go up.

guidanoli:
Here is the difference in gas costs from `main` to `feature/input-idx-uint64`.
Cost-wise, I don't think it's too relevant of a cost change.
The greatest changes, in absolute terms, are on deployments, which would cost 1 to 2 dollars more.

| Contract | Function | Min | Avg | Max |
| :- | :- | :-: | :-: | :-: |
| Application | executeVoucher | 804 (12.41%) | 877 (1.63%) | 927 (0.94%) |
| Application | getInputRelays | -22 (-0.07%) | -22 (-0.06%) | -22 (-0.04%) |
| Application | validateNotice | 878 (2.83%) | 878 (2.55%) | 878 (2.31%) |
| Application | wasVoucherExecuted | 253 (41.00%) | 253 (19.72%) | 253 (9.67%) |
| ApplicationFactory | | | | 39642 (2.55%) |
| ApplicationFactory | calculateApplicationAddress | 573 (0.67%) | 573 (0.67%) | 573 (0.67%) |
| ApplicationFactory | newApplication(address,address,address[],address,bytes32) | 39664 (1.07%) | 39664 (1.07%) | 39664 (1.07%) |
| ApplicationFactory | newApplication(address,address,address[],address,bytes32,bytes32) | 39703 (0.74%) | 13282 (0.00%) | -39561 (-0.00%) |
| ApplicationTest | | | | 127441 (1.27%) |
| ApplicationTest | calculateInputIndex | 221 (28.01%) | 221 (28.01%) | 221 (28.01%) |
| Authority | | | | -212787 (-100.00%) |
| Authority | getEpochHash | 449 (60.19%) | 449 (60.19%) | 449 (60.19%) |
| Authority | owner | -33 (-9.57%) | -33 (-9.57%) | -33 (-9.57%) |
| Authority | submitClaim | | 505 (4.54%) | 757 (2.77%) |
| AuthorityFactory | | | | 48450 (11.83%) |
| AuthorityFactory | calculateAuthorityAddress | 621 (13.57%) | 621 (13.57%) | 621 (13.57%) |
| AuthorityFactory | newAuthority(address) | 48492 (19.64%) | 48492 (19.64%) | 48492 (19.64%) |
| AuthorityFactory | newAuthority(address,bytes32) | 48534 (19.64%) | 167 (0.00%) | -48199 (-0.00%) |
| Quorum | | | | 63262 (21.01%) |
| Quorum | numOfValidators | 45 (24.86%) | 45 (24.86%) | 45 (24.86%) |
| Quorum | validatorById | -22 (-4.21%) | -74 (-12.89%) | -2022 (-80.17%) |
| Quorum | validatorId | | -95 (-7.96%) | |
| QuorumFactory | | | | 63258 (10.17%) |
| QuorumFactory | calculateQuorumAddress | 892 (11.75%) | 892 (11.75%) | 892 (11.75%) |
| QuorumFactory | newQuorum(address[]) | 63294 (3.97%) | 63294 (3.97%) | 63294 (3.97%) |
| QuorumFactory | newQuorum(address[],bytes32) | 63385 (16.65%) | 108 (0.00%) | -63169 (-0.00%)

onchain/rollups/contracts/dapp/Application.sol
```diff
@@ -123,7 +125,7 @@ contract Application is
         // reverts if proof isn't valid
         proof.validity.validateVoucher(destination, payload, epochHash);
 
-        uint256 outputIndexWithinInput = proof.validity.outputIndexWithinInput;
+        uint64 outputIndexWithinInput = proof.validity.outputIndexWithinInput;
```
ZzzzHui, line 128: we could also keep this as `uint256` and call `toUint64()` when emitting `VoucherExecuted`, like you did for `inputIndex`

onchain/rollups/contracts/library/LibProof.sol
```diff
@@ -10,7 +10,7 @@ library LibProof {
         Proof calldata proof
     ) internal pure returns (uint256 inputIndex) {
         inputIndex =
-            proof.inputRange.firstIndex +
+            uint256(proof.inputRange.firstIndex) +
```
ZzzzHui, line 13: we can omit this cast

onchain/rollups/contracts/dapp/Application.sol
guidanoli: We could, but it would be super weird to cast a `uint64` to a `uint256` then call `toUint64` on it.

onchain/rollups/contracts/library/LibProof.sol
guidanoli: If we omit the cast, it will do addition on `uint64` values, which might be prone to overflows.

ZzzzHui: It only overflows if the input index is greater than 2^64, which will not happen right?

guidanoli: Yes, one will not be able to prove anything with an input index greater than or equal to $2^{64}$, because there is no `InputRange` that would contain it. The key difference is that, from the perspective of the user, the error would be different.
If we do the addition in...
- `uint256`, the error would be `InputIndexOutOfRange`
- `uint64`, the error would be `Panic` with error code of an arithmetic overflow
I think the first is easier for the user to understand.

ZzzzHui (approved):

PR #216, opened 2024-01-22
refactor: use `abi.encode` and remove dead code

### LINKED ISSUE #215: Use `abi.encode` for fixed-size values
## 📚 Context

There are places in our code base that use `abi.encodePacked` for fixed-size values.
This is not necessary, as `abi.encode` works just like `abi.encodePacked` for those values.
Plus, `abi.encode` is shorter.

## ✔️ Solution

Replace `abi.encodePacked` with `abi.encode` in those cases.

guidanoli:
Closes #215

### DIFF
--- a/onchain/rollups/test/foundry/dapp/Application.t.sol
+++ b/onchain/rollups/test/foundry/dapp/Application.t.sol
@@ -977,7 +977,7 @@ contract ApplicationTest is TestBase {
     ) internal pure returns (bytes32) {
         return
             keccak256(
-                abi.encodePacked(
+                abi.encode(
                     validity.outputsEpochRootHash,
                     validity.machineStateHash
                 )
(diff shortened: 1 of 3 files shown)


onchain/rollups/test/foundry/dapp/Application.t.sol
```diff
@@ -195,7 +195,7 @@ contract ApplicationTest is TestBase {
         _validateNotice(notice, proof);
 
         // reverts if notice is incorrect
-        bytes memory falseNotice = abi.encodePacked(bytes4(0xdeaddead));
+        bytes memory falseNotice = abi.encode(bytes4(0xdeaddead));
```
ZzzzHui, line 198: These 2 behaviors are different. But the result is the same because they are both false notices.

ZzzzHui (dismissed):

onchain/rollups/test/foundry/dapp/Application.t.sol
guidanoli: Ooops, you're right. Let me revert this line.

ZzzzHui: Ok. Either way was ok :)

ZzzzHui (approved):

PR #217, opened 2024-01-23
refactor!: move `InputRange` inside `OutputValidityProof`

### LINKED ISSUE #213: Move `InputRange` field to inside `OutputValidityProof` struct
## 📚 Context

As of v1.2.0, in order to execute a voucher, one has to provide the voucher fields (destination and payload) and a proof structure. This proof structure contains an `OutputValidityProof` field and a proof context blob.

With #42, the specific outputs fields would be abstracted away, and the user would provide the output blob instead, to be decoded by the contract itself.

With #48, the proof context blob is no longer necessary, and we, instead, use an `InputRange` structure to index claims for a given application.

As a result, the signature of `executeVoucher` would go from...

```solidity
struct Proof {
  OutputValidityProof validity;
  bytes context;
}

function executeVoucher(address destination, bytes calldata payload, Proof calldata proof) external;
```

... to...

```solidity
struct Proof {
  OutputValidityProof validity;
  InputRange inputRange;
}

function executeOutput(bytes calldata output, Proof calldata proof) external;
```

In the server manager gRPC interface, however, these two fields are stored separately. The node doesn't really care about their contents specifically, and would like to look at them as opaque.

## ✔️ Solution

After discussing with the machine and node teams, we agreed to add an `InputRange` field to the `OutputValidityProof` structure. This would also be reflected on the gRPC interfaces used by the server manager, for example.

As a result, we can remove the `Proof` structure from the code base, and use `OutputValidityProof` instead.

```solidity
struct OutputValidityProof {
  InputRange inputRange;
  // the other fields...
}

function executeOutput(bytes calldata output, OutputValidityProof calldata proof) external;
```

guidanoli:
Closes #213


onchain/rollups/test/foundry/dapp/Application.t.sol
```diff
@@ -944,11 +950,7 @@ contract ApplicationTest is TestBase {
         for (uint256 i; i < proofs.length; ++i) {
             LibServerManager.Proof memory proof = proofs[i];
             if (proof.proves(outputEnum, inputIndexWithinEpoch, outputIndex)) {
-                return
-                    Proof({
-                        validity: _convert(proof.validity),
-                        inputRange: inputRange
-                    });
+                return _convert(proof.validity, inputRange);
```
ZzzzHui, line 953: `LibServerManager` would be updated after the server manager updates its response content, right?

ZzzzHui (approved):

onchain/rollups/test/foundry/dapp/Application.t.sol
guidanoli: Yes, there are just artifacts that lets us test the new output validation scheme using the old machine.

ZzzzHui (approved):

PR #219, opened 2024-01-23
chore: organize changesets

guidanoli:


onchain/rollups/.changeset/clean-gorillas-occur.md
```diff
@@ -0,0 +1,9 @@
+---
+"@cartesi/rollups": major
+---
+
+Modified the `CanonicalMachine` library:
+
+-   Renamed the `INPUT_MAX_SIZE` constant as `INPUT_PAYLOAD_MAX_SIZE`.
```
ZzzzHui, line 7: I think this is an intermediary step. Maybe we can omit it instead of having 2 change logs that make and undo the change.

onchain/rollups/.changeset/chilly-cherries-join.md
```diff
@@ -0,0 +1,7 @@
+---
+"@cartesi/rollups": major
+---
+
+Modified the ERC-20 deposit input.
```
ZzzzHui, line 5: This is minor, but should this sentence end with `:`?

guidanoli: Yes, thanks. :-)

onchain/rollups/.changeset/clean-gorillas-occur.md
guidanoli: Whenever a developer makes a change, they should hunt for references to it in the codebase, which may need to be removed, adapted, or improved. For example, if you remove the `withdrawEther` function, you have to remove some test cases, change the `README` part about withdrawing Ether, etc. Additionally, they should also look for `.changeset` files. If there is already a `.changeset` file related to the change, they should edit it. Otherwise, they should create a new one.

ZzzzHui (approved):

onchain/rollups/.changeset/clean-gorillas-occur.md
ZzzzHui: From this angle, changelog does have an advantage over changeset, in the sense that it doesn't need to jump between files

PR #222, opened 2024-01-24
include chain id in `EvmAdvance` input

ZzzzHui:


ZzzzHui:
According to [EIP 1344](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1344.md#specification), chain ID is a 256-bit value, but some toolings may limit the chain ID in the implementation. Like in our stack, chain ID is limited to be less than 2^64-1. So we use `uint64` in the fuzzy tests.

pedroargento (dismissed):

guidanoli (approved):

onchain/rollups/.changeset/shaggy-queens-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cartesi/rollups": major
+---
+
+Include chain id in `EvmAdvance` input.
```
guidanoli, line 5:
```suggestion
Include chain ID in `EvmAdvance` input.
```

guidanoli (approved):

PR #228, opened 2024-01-25
refactor!: remove application address relay

### LINKED ISSUE #225: Remove `ApplicationAddressRelay` contract
## 📚 Context

After the issue ["Include the DApp address in the EvmAdvance input"](https://github.com/cartesi/rollups-contracts/issues/144), the contract `ApplicationAddressRelay` would be of no use. So it should be removed. Also need to update documentation and Mermaid graph.

guidanoli:
Closes #225


ZzzzHui:
[Here](https://github.com/cartesi/rollups-contracts/blob/bbe4b4c429a749e3385a5d9d6f38e17c6ffdb518/onchain/rollups/contracts/common/InputEncoding.sol#L13) in the library `InputEncoding`, it says `such as portals and relays.` Now there's only `portals`. So we should update this as well. 
Also, now that we've removed address relay, I'm thinking about whether we need to introduce the new term `relay` at all, or should we keep using the word `portal`? On the other hand, the word `relay` still makes sense and is more general. What do you think?

guidanoli:
> Also, now that we've removed address relay, I'm thinking about whether we need to introduce the new term relay at all, or should we keep using the word portal?

I agree. Let's call them portals in another PR.

PR #240, opened 2024-01-30
rename input relay as portal

ZzzzHui:


guidanoli (commented):
In some cases, portals are just documented as "The portals", but I think that's too little information. I'd add the word "supported" to give the idea of there being portals that the application does not support.

guidanoli:
The changeset files need to be adjusted. See https://github.com/cartesi/rollups-contracts/pull/241.

guidanoli (approved):

PR #241, opened 2024-01-30
chore: adjust changeset files

guidanoli:
- `IInputRelay` -> `IPortal`
- `InputRelay` -> `Portal`
- `inputRelays` -> `portals`
- `getInputRelays` -> `getPortals`
- `DAppAddressRelay` was removed, not renamed

### DIFF
--- a/onchain/rollups/.changeset/fast-tips-push.md
+++ b/onchain/rollups/.changeset/fast-tips-push.md
@@ -1,6 +1,0 @@
----
-"@cartesi/rollups": major
----
-
-Renamed `IInputRelay` as `IPortal` and `InputRelay` as `Portal`.
-Moved those 2 contracts to the folder `portals`.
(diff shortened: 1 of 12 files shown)


ZzzzHui (approved):
Thanks for arranging the changset files. 
Still not sure if we should, re-consider using changelog than changeset? Do you think changeset poses more difficulties than bringing convenience?

guidanoli:
> Still not sure if we should, re-consider using changelog than changeset? Do you think changeset poses more difficulties than bringing convenience?

I think [changesets](https://github.com/changesets/changesets) still bring us some benefits:

- They avoid conflicts in PRs, by spreading out changes across multiple files
- They simplify the process of versioning packages and publishing releases and pre-releases, through a handy-dandy CLI tool and GitHub Action

guidanoli:
The necessity of keeping changes up-to-date exists with or without changesets. The difference is that changes are spread out across multiple files, so you may need to do full-text search (`git grep` works wonders).

PR #245, opened 2024-02-01
ci: deploy documentation

### LINKED ISSUE #123: Build docs in the workflow
## 📚 Context

Currently, docs are built and formatted after each release.
It would be nice, however, to have the docs be built every time a new commit is pushed to the main branch.
This way, we can detect improvements in the docs early in the development process.

## ✔️ Solution

One way to achieve this is to let GitHub Actions build the docs.
I suggest we use `forge doc` so that we add no extra dependency.
This command outputs a directory named `docs`, which contains browser-ready HTML.
We can then deploy these files to GitHub Pages using some already available GitHub Actions.
I've already done something like this in the past for a university course: [See here](https://github.com/guidanoli/catan-lua/blob/main/.github/workflows/docs.yml).

## 📈 Subtasks

- [ ] Add job that build docs
- [ ] Add job that deploys docs to GitHub Pages

guidanoli:
Closes #123

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,11 @@ yarn proofs:update
 
 ## 📚 Documentation
 
-ℹ️ Check the [official Cartesi Rollups documentation website](https://docs.cartesi.io/cartesi-rollups/overview/).
+🚀 Smart contract documentation is kept up-to-date and deployed [**here**](https://cartesi.github.io/rollups-contracts).
 
-For an in-depth view of the on-chain architecture, we invite you to take a look at the [`CONTRACTS.md`](./CONTRACTS.md) file.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -66,9 +66,11 @@ yarn proofs:update
 
 ## 📚 Documentation
 
-ℹ️ Check the [official Cartesi Rollups documentation website](https://docs.cartesi.io/cartesi-rollups/overview/).
+🚀 Smart contract documentation is kept up-to-date and deployed [**here**](https://cartesi.github.io/rollups-contracts).
```
ZzzzHui, line 69: is this the correct url?

guidanoli: Yes, it will be live as soon as this gets merged and runs for the first time.

pedroargento (approved):

PR #248, opened 2024-02-21
Refactor tests

guidanoli:


test/foundry/dapp/Application.t.sol
ZzzzHui, line 1: We could remove the _mockConsensus function since it’s no longer used :)

guidanoli: Thanks! Removed, along with `_calculateEpochHash`.

ZzzzHui (approved):
not sure why the CI tests failed. All tests passed on my end. So maybe re-run the CI

guidanoli:
> not sure why the CI tests failed. All tests passed on my end. So maybe re-run the CI

Yeah, very weird. In the CI run, the error is a panic `0x32` which signals and out-of-bounds access to an array, which is hard to debug on its own... So, I took a look around `Application.t.sol` for array accesses, and I think it could have come from `_getOutput`, so I added a check there, which raises an `OutputNotFound` custom error instead of panicking.

guidanoli:
Oops, accidentally re-added `_mockConsensus` and `_calculateEpochHash`.
Removed them once again.

ZzzzHui:
Hmmm. Still same error

guidanoli:
> Hmmm. Still same error

Yep, I think I managed to pinpoint the error.

guidanoli:
Now everything is fixed.

ZzzzHui (approved):

guidanoli:
- Test executing voucher for payable functions
- Test validating new type of output (new function selector)
- Test executing empty output reverts
- Test executing notice reverts
- Test executing new type of output reverts

test/foundry/dapp/Application.t.sol
```diff
@@ -455,6 +458,35 @@ contract ApplicationTest is ERC165Test {
         _app.executeOutput(output, proof);
     }
 
+    function testEtherReceiver() public {
+        uint256 value = address(this).balance;
+        (bool success, ) = address(_etherReceiver).call{value: value}("");
+        assertFalse(success, "expected EtherReceiver to revert");
+    }
```
ZzzzHui, line 465: Not sure if we need this test?

test/foundry/dapp/Application.t.sol
```diff
@@ -46,18 +47,21 @@ contract ApplicationTest is ERC165Test {
         HelloWorld,
         MyOutput,
         ETHTransfer,
+        ETHMint,
```
ZzzzHui, line 50: we could consider renaming these 2 outputs as: `ETHTransferEOA` `ETHTransferContract`

guidanoli: Ok, I'll rename them.

guidanoli: How about `ETHTransferNoPayload` and `ETHTransferWithPayload`?
I think `ETHTransferEOA` might be misleading, because the voucher `(recipient,value,"")` can be used for both EOAs and contract recipients. Likewise, a voucher `(recipient,value,payload)` (where `payload` is not empty) would also succeed if `recipient` were an EOA.

guidanoli: How about `ETHTransfer` and `PayableFunction`?

guidanoli: Because `ETHTransfer := (recipient,value,"")` can be used both for EOAs and contracts, so saying it's `ETHTransferEOA` might be misleading.

test/foundry/dapp/Application.t.sol
guidanoli: Removed.

test/foundry/dapp/Application.t.sol
ZzzzHui: In our setting `_recipient = LibBytes.hashToAddress("recipient");` is basically an EOA right?
But I do like `ETHTransfer` and `PayableFunc` much more, so let's do this ones :)

guidanoli: Yes, because `_recipient` has no code by default, but we could inject code into it to test the scenario where the recipient is a contract.

guidanoli:
- Adjusted portal integration tests to expect `addInput` calls and not use holder contracts, since token transfers work perfectly well with EOA recipients.
- Cover ERC-1155 batch transfers with 0 and 1 tokens
- Fix `vm.expectEmit` missing emitter address

ZzzzHui (approved):

PR #257, opened 2024-04-04
feat: add `block.prevrandao` to input metadata

### LINKED ISSUE #256: Add `block.prevrandao` to input metadata
## 📚 Context

After the Merge, the header of every block contains a field called `prevrandao`, a random number generated by the beacon chain that is used by the proof-of-stake to randomly elect block proposers in the beginning of every epoch.

Applications can use this number as a source of randomness if they don't intend to handle large sums of assets. For large TVLs, we do not recommend using `block.prevrandao`. See [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399) for technical details, and security tips for application developers.

## ✔️ Solution

We can add `prevrandao` to every input as metadata.
This would be reflected in the current architecture as an extra parameter in the `EvmInspect` function.
In Solidity, we'll call it `prevRandao` so that application developers can quickly find resources based on the real name.

guidanoli:
Closes #256

### DIFF
--- a/contracts/inputs/InputBox.sol
+++ b/contracts/inputs/InputBox.sol
@@ -1,7 +1,7 @@
 // (c) Cartesi and individual authors (see AUTHORS)
 // SPDX-License-Identifier: Apache-2.0 (see LICENSE)
 
-pragma solidity ^0.8.13;
+pragma solidity ^0.8.18;
 
 import {IInputBox} from "./IInputBox.sol";
 import {CanonicalMachine} from "../common/CanonicalMachine.sol";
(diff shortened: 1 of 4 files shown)


contracts/inputs/InputBox.sol
```diff
@@ -1,7 +1,7 @@
 // (c) Cartesi and individual authors (see AUTHORS)
 // SPDX-License-Identifier: Apache-2.0 (see LICENSE)
 
-pragma solidity ^0.8.13;
+pragma solidity ^0.8.18;
```
guidanoli, line 4: The `block.prevrandao` global variable was introduced in [0.8.18](https://github.com/ethereum/solidity/releases/tag/v0.8.18).

PR #264, opened 2024-04-11
Refactor output validation tests

### LINKED ISSUE #266: Fix proof verification for Output Unification
## 📚 Context

As of SDK v1, the epoch hash calculation was grounded on the memory space of the machine, which has 8-byte words. For SDK v2, the Merkle trees will be constructed on top of 32-byte words, namely the output hashes. This means that we need to update the on-chain proof verification algorithm to work under these new premises.

Merkle proofs are currently done by the `MerkleV2` library from `solidity-util`, which has several optimizations for working with 8-byte leaves, such as pre-computed pristine hashes. This library, however, is not suitable for our needs.

## ✔️ Solution

We need to develop a new library for working with Merkle trees with 32-byte leaves.
We should thoroughly test this library with concrete and fuzzy tests.
For Output Unification Part 1, we'll use this library but not `MerkleV2`.
As for Output Unification Part 2, we'll use this library AND `MerkleV2`.
It will be necessary then to use `MerkleV2` because the epoch hash will be equal to the machine state hash, and we'll have to prove that the output Merkle tree root hash is contained in the memory space of the machine, which contains 8-byte words.

guidanoli:
Closes #266 and #265


guidanoli:
## Gas analysis

It seems that the gas cost of validating (and, therefore, executing) outputs has decreased significantly with these changes. Maybe it has something to do with the fact that we're now calling `internal` library functions (which translate to `JUMP` instructions) rather than calling `external` library functions (which translate to `DELEGATECALL` instructions, which are more expensive)?

| | Minimum | Average | Median | Maximum |
| :-: | :-: | :-: | :-: | :-: |
| Before | 32662 | 33509 | 32687 | 44163 |
| After | 16464 | 18009 | 16554 | 22965 |
| Difference | -49.6% | -46.2% | -49.4% | -48.0% |

guidanoli:
I've also broken long lines in `CONTRACTS.md` so that future diffs (which are often line-oriented) are easier to read.

contracts/library/LibMerkle32.sol
ZzzzHui, line 1: The creation of merkle tree is not onchain right? We can move the creation part to the test folder if it's only used there.
Even for the verification part, maybe we should use the OpenZeppelin `MerkleProof` instead. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/MerkleProof.sol
What do you think?

guidanoli: > The creation of merkle tree is not onchain right? We can move the creation part to the test folder if it's only used there.

So, although the `merkleRoot` and `siblings` functions are only used by the unit tests, they also serve to document how the inputs to the `merkleRootAfterReplacement` function should be computed from the leaves of a Merkle tree. So, I would leave these functions in the same library.

guidanoli: > Even for the verification part, maybe we should use the OpenZeppelin MerkleProof instead.

OpenZeppelin's Merkle trees are not the same as the Merkle trees generated by the `libmct` library. In particular, OpenZeppelin's siblings are sorted before hashing. See [`Hashes.sol`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/Hashes.sol).

pedroargento (approved):

ZzzzHui (approved):

PR #268, opened 2024-04-12
Add convenience application factory for self-hosting

tuler:
This is a convenient application factory that calls the IAuthorityHistoryPairFactory followed by the CartesiDAppFactory, to simplify the deployment of application that will be validated by an authority.

It's being used by the sunodo deployment system at https://sunodo.io/deploy

With the move from Sunodo CLI to Cartesi CLI we will probably have to move this contract away from the sunodo repo.


tuler:
There is a way to keep this contract under the sunodo context, if we assume that it's only for the sunodo webapp, as a convenience.

And make the Cartesi CLI not use it, and send two transactions on application startup.

pedroargento:
The ` IAuthorityHistoryPairFactory` is not needed in newer versions. The consensus now inherits the history functionalities from `AbstractConsensus`  and there is no need for the `History` contract anymore.

tuler:
Ok, but this is for V1

guidanoli:
I've rebased `release/1.4.x` on top of `release/1.3.x` because of the `1.3.1` release.

guidanoli:
I've made some small adjustments, and added tests.
We can squash everything once it's merge time.

ZzzzHui (approved):

guidanoli:
@tuler Do you approve these changes? I can't tag you as reviewer since you are the one who created this PR.

guidanoli (approved):

PR #271, opened 2024-04-19
safe erc20 delegate

ZzzzHui:
I call it delegate and create the folder `delegate`.
Let me know if you think it should be named something else

### DIFF
--- a/.changeset/lucky-elephants-play.md
+++ b/.changeset/lucky-elephants-play.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/rollups": minor
+---
+
+Add a contract for safe ERC20 transfers. This can be used by delegatecall vouchers.
--- a/test/foundry/dapp/Application.t.sol
+++ b/test/foundry/dapp/Application.t.sol
@@ -19,9 +19,11 @@ import {InputBox} from "contracts/inputs/InputBox.sol";
 import {InputRange} from "contracts/common/InputRange.sol";
 import {OutputValidityProof} from "contracts/common/OutputValidityProof.sol";
 import {Outputs} from "contracts/common/Outputs.sol";
+import {SafeERC20Transfer} from "contracts/delegatecall/SafeERC20Transfer.sol";
 
 import {IERC1155Receiver} from "@openzeppelin/contracts/token/ERC1155/IERC1155Receiver.sol";
 import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import {IERC20Errors, IERC721Errors} from "@openzeppelin/contracts/interfaces/draft-IERC6093.sol";
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import {IERC721Receiver} from "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
@@ -35,8 +37,6 @@ import {SimpleERC20} from "../util/SimpleERC20.sol";
 import {SimpleERC721} from "../util/SimpleERC721.sol";
 import {ExternalLibMerkle32} from "../library/LibMerkle32.t.sol";
 
-import "forge-std/console.sol";
-
 contract ApplicationTest is ERC165Test {
     using LibEmulator for LibEmulator.State;
     using ExternalLibMerkle32 for bytes32[];
@@ -48,6 +48,7 @@ contract ApplicationTest is ERC165Test {
     IERC721 _erc721Token;
     IInputBox _inputBox;
     IPortal[] _portals;
+    SafeERC20Transfer _safeERC20Transfer;
     LibEmulator.State _emulator;
     address _appOwner;
     address _authorityOwner;
@@ -302,8 +303,8 @@ contract ApplicationTest is ERC165Test {
         _appContract.executeOutput(output, proof);
     }
 
-    function testExecuteErc20TransferVoucher() external {
-        string memory name = "Erc20TransferVoucher";
+    function testExecuteERC20TransferVoucher() external {
+        string memory name = "ERC20TransferVoucher";
         bytes memory output = _getOutput(name);
         OutputValidityProof memory proof = _getProof(name);
 
@@ -353,8 +354,8 @@ contract ApplicationTest is ERC165Test {
         _appContract.executeOutput(output, proof);
     }
 
-    function testExecuteErc721TransferVoucher() external {
-        string memory name = "Erc721TransferVoucher";
+    function testExecuteERC721TransferVoucher() external {
+        string memory name = "ERC721TransferVoucher";
         bytes memory output = _getOutput(name);
         OutputValidityProof memory proof = _getProof(name);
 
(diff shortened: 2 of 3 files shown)


guidanoli:
Why not make it a library? They won't be able to access their storage anyway when they are called by the application contract.

guidanoli:
I'd also rename it to something like `LibERC20Transfer`

ZzzzHui:
> Why not make it a library? It won't be able to access its storage anyway when called by the application contract via `DELEGATECALL`.

It's basically the same in this case to be a contract and library. But `library` is indeed a better name, so I will change `contract` to `library`.

contracts/delegate/ERC20Delegate.sol
```diff
@@ -0,0 +1,23 @@
...
+import {SafeERC20, IERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+contract ERC20Delegate {
+    using SafeERC20 for IERC20;
+
+    function safeTransfer(IERC20 token, address to, uint256 value) external {
+        token.safeTransfer(to, value);
+    }
+
+    function safeTransferFrom(
```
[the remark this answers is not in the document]
ZzzzHui: I'm not too sure about this actually. `transferFrom` is quite a common functionality. Should we leave the use cases to dapp developers?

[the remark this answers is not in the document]
guidanoli: The `transferFrom` function should only be needed if the `from` argument is not the application contract address. Most applications, however, would only need the `transfer` function to withdraw assets locked in the application contract.

guidanoli:
> It's basically the same in this case to be a contract and library. But `library` is indeed a better name, so I will change `contract` to `library`.

It's not just about naming. In Solidity, they have different semantics. For example, state variables cannot be declared. This avoids programming errors. So, if the contract is going to be called via `DELEGATECALL`, it's good practice to declare it as a library.

test/foundry/library/LibERC20Transfer.t.sol
```diff
@@ -0,0 +1,95 @@
...
+            )
+        );
+        _token.safeTransfer(to, value);
+    }
+
+    function testNormalToken(uint256 supply, address to, uint256 value) public {
+        vm.assume(to != address(0));
+        supply = bound(supply, 1, type(uint256).max - 1);
+        value = bound(value, 0, supply);
+        _token = new NormalToken(address(this), supply);
```
guidanoli, line 63: You could define a `_tokenOwner` state variable and assign it to `_newAddr` in the `setUp` function.

guidanoli (commented):
Missing `changeset` file.

test/foundry/library/LibERC20Transfer.t.sol
ZzzzHui: Yes. Although since this var is only used in this test. I will place the var in the current test instead of `setUp` function :)

ZzzzHui:
I think we should keep these delegate libraries in a separate folder, rather than mixing together with existing internal libraries. I think we could put them in the folder called `delegate`. What do you think?

guidanoli (commented):
Also, would be nice, if possible, to add test cases to `Application.t.sol` to test the `DELEGATECALL` voucher feature. :-)

guidanoli:
> I think we should keep these delegate libraries in a separate folder, rather than mixing together with existing internal libraries. I think we could put them in the folder called `delegate`. What do you think?

The word "delegate" can be confusing for those familiar with "delegated proof-of-staking". Maybe `delegatecall`?

ZzzzHui:
> > I think we should keep these delegate libraries in a separate folder, rather than mixing together with existing internal libraries. I think we could put them in the folder called `delegate`. What do you think?
> 
> The word "delegate" can be confusing for those familiar with "delegated proof-of-staking". Maybe `delegatecall`?

Yes. `delegatecall` will work.

> Also, would be nice, if possible, to add test cases to Application.t.sol to test the DELEGATECALL voucher feature. :-)

I agree that would be better.

test/foundry/util/ContractERC20Transfer.sol
guidanoli, line 1: I don't understand the necessity of this contract.
Why not use the library?

ZzzzHui: In order to `delegatecall`, we need the address of the library. But I didn't see a way to create a library instance, which means we don't know the address of the library. Thus, I created a contract as a caller to library. 
Let me know if I miss something :)

guidanoli: I see, it's not so easy to deploy libraries in Solidity on their own.
Let's use contracts then, I think it will be easier then.

test/foundry/dapp/Application.t.sol
```diff
@@ -541,6 +593,23 @@ contract ApplicationTest is ERC165Test {
         _finishInput();
         _finishEpoch();
 
+        _nameOutput(
+            "ERC20DelegateCallVoucher",
+            _addOutput(
+                _encodeDelegateCallVoucher(
+                    address(_contractERC20Transfer),
+                    abi.encodeWithSignature(
```
guidanoli, line 601: Prefer `abi.encodeCall` over `abi.encodeWithSignature`, as the former allows the compiler to make type checks on the arguments of the call.

guidanoli (commented):
Let's convert the library into a contract then, since it will be easier to test with Forge. For consistency, let's rename it to `SafeERC20Transfer` or just `ERC20Transfer`. With this, we can remove `ContractERC02Transfer`.

Also, I'd remove `LibERC20Transfer.t.sol` and add individual test cases to `Application.t.sol`, since it will be harder or less natural to do explicit delegate calls instead of the `using <Library> for <Type>` syntax.

test/foundry/dapp/Application.t.sol
ZzzzHui: Yes `abi.encodeCall` is preferred. But we need an interface to be used by `encodeCall`, right?

ZzzzHui: Yes. Should use `abi.encodeCall` instead.

.changeset/lucky-elephants-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cartesi/rollups": minor
+---
+
+Add a library for safe ERC20 transfers. This can be used by delegatecall vouchers.
```
guidanoli, line 5: It is now a contract rather than a library.

test/foundry/dapp/Application.t.sol
```diff
@@ -49,6 +49,7 @@ contract ApplicationTest is ERC165Test {
     IInputBox _inputBox;
     IPortal[] _portals;
     LibEmulator.State _emulator;
+    SafeERC20Transfer _safeERC20Transfer;
```
guidanoli, line 52: I'd switch these two.
Note that we're first defining contracts (application, input box, portals, etc.) and then variables like emulator state, application owner, etc. So `SafeERC20Transfer` should be grouped with the contracts.

test/foundry/dapp/Application.t.sol
```diff
@@ -425,6 +426,78 @@ contract ApplicationTest is ERC165Test {
...
+            abi.encodeWithSelector(
+                IERC20Errors.ERC20InsufficientBalance.selector,
+                address(_appContract),
+                _erc20Token.balanceOf(address(_appContract)),
+                _transferAmount
+            )
+        );
+        _appContract.executeOutput(output, proof);
+
+        // fund dapp
```
guidanoli, line 452: It's self-evident that we're funding the application contract, so we don't need to comment that.

guidanoli (approved):

PR #273, opened 2024-04-23
ENS libs for delegatecall vouchers

ZzzzHui:


contracts/delegatecall/LibEtherForENS.sol
```diff
@@ -0,0 +1,25 @@
...
+}
+
+abstract contract ENS {
+    function resolver(bytes32 node) public view virtual returns (Resolver);
+}
+
+library LibEtherForENS {
+    function etherForENS(bytes32 node, uint256 value, bytes memory payload) external{
+        // address of The ENS registry
+        ENS ens = ENS(0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e);
```
ZzzzHui, line 17: This address is the same for the networks that ENS root registry has been deployed [so far](https://docs.ens.domains/learn/deployments). It will likely be the same for other future networks as well. But if that's not the case, then this address should be modify accordingly to deploy this library onto that network.

guidanoli: I'd be more comfortable with this address being provided in the constructor. This is, considering we convert this library into a contract. Then, we can provide the correct address on the deployment script depending on the chain.

guidanoli: One nice experiment would be to see if the contract can access `immutable` variables such as the ENS registry contract even when it's called via `delegatecall`. My intuition tells me that immutable variables are written to byte code at contract creation time.

contracts/delegatecall/LibEtherForENS.sol
```diff
@@ -0,0 +1,25 @@
...
+
+pragma solidity ^0.8.20;
+
+abstract contract Resolver {
+    function addr(bytes32 node) public view virtual returns (address);
+}
+
+abstract contract ENS {
+    function resolver(bytes32 node) public view virtual returns (Resolver);
+}
```
guidanoli, line 12: Maybe we could add `@ensdomains/ens-contracts` as a dependency and import these definitions from there?

contracts/delegatecall/LibEtherForENS.sol
```diff
@@ -0,0 +1,25 @@
...
+library LibEtherForENS {
+    function etherForENS(bytes32 node, uint256 value, bytes memory payload) external{
+        // address of The ENS registry
+        ENS ens = ENS(0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e);
+
+        Resolver resolver = ens.resolver(node);
+        address resolution = resolver.addr(node);
+
+        (bool success, bytes memory returndata) = resolution.call{value: value}(payload);
+        require(success, string(returndata));
```
guidanoli, line 23: Please prefer using the following function instead for propagating errors.

https://github.com/cartesi/rollups-contracts/blob/2ab9e2388eed4f9f56b0cd2a45be320d5d3e1001/contracts/library/LibError.sol#L9

contracts/delegatecall/LibEtherForENS.sol
```diff
@@ -0,0 +1,25 @@
...
+    function resolver(bytes32 node) public view virtual returns (Resolver);
+}
+
+library LibEtherForENS {
+    function etherForENS(bytes32 node, uint256 value, bytes memory payload) external{
+        // address of The ENS registry
+        ENS ens = ENS(0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e);
+
+        Resolver resolver = ens.resolver(node);
+        address resolution = resolver.addr(node);
```
guidanoli, line 20: Why not `receiver` instead of `resolution`?

contracts/delegatecall/LibEtherForENS.sol
```diff
@@ -0,0 +1,25 @@
...
+abstract contract Resolver {
+    function addr(bytes32 node) public view virtual returns (address);
+}
+
+abstract contract ENS {
+    function resolver(bytes32 node) public view virtual returns (Resolver);
+}
+
+library LibEtherForENS {
+    function etherForENS(bytes32 node, uint256 value, bytes memory payload) external{
```
guidanoli, line 15: Given that we're allowing a `payload` to be passed, this is more like an arbitrary low-level call to an address through its ENS name. Maybe rename this library to `LibCallToENS`?

contracts/delegatecall/LibEtherForENS.sol
ZzzzHui: Yes we can. It's a different perspective. `recipient` is viewed from the transfer, `resolution` is viewed from the ENS. I will rename it to `recipient`.

contracts/delegatecall/LibEtherForENS.sol
ZzzzHui: Yeah. `immutable` variables should be in the bytecode during deployment. I have done an experiment to confirm that ✅

contracts/delegatecall/CallToENS.sol
```diff
@@ -0,0 +1,90 @@
...
+import {IERC1155} from "@openzeppelin/contracts/token/ERC1155/IERC1155.sol";
+
+import {LibError} from "../library/LibError.sol";
+
+contract CallToENS {
+    using LibError for bytes;
+    using SafeERC20 for IERC20;
+
+    ENS immutable _ens;
+    constructor(ENS ens) {
```
guidanoli, line 20: I'd suggest adding an empty line in-between these two for readability.

contracts/delegatecall/CallToENS.sol
```diff
@@ -0,0 +1,90 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+pragma solidity ^0.8.20;
+
+import {ENS} from "@ensdomains/ens-contracts/contracts/registry/ENS.sol";
+import {AddrResolver} from "@ensdomains/ens-contracts/contracts/resolvers/profiles/AddrResolver.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
```
guidanoli, line 8: I'd add an empty line to separate imports from different packages.

contracts/delegatecall/CallToENS.sol
```diff
@@ -0,0 +1,90 @@
...
+    function sendCallToENS(
+        bytes32 node,
+        uint256 value,
+        bytes memory payload
+    ) external {
+        address recipient = resolveENS(node);
+
+        (bool success, bytes memory returndata) = recipient.call{value: value}(
+            payload
+        );
```
guidanoli, line 33: You can first declare these variables, and then assign them, so that the line fits in 80 characters.

guidanoli (commented):
If we decide to take the "all-in-one" approach, I think we could think of a better name for the contract, since we're not really calling addresses by their ENS names in all the cases.

Also, one nice thing I noticed is that this library frees the back-end from providing the application contract address as argument. This information is retrieved at runtime via `address(this)`.

contracts/delegatecall/CallToENS.sol
ZzzzHui: It's not more than 80 chars in its current way actually

ZzzzHui:
> If we decide to take the "all-in-one" approach, I think we could think of a better name for the contract, since we're not really calling addresses by their ENS names in all the cases.

How about `CallWithENS`?

contracts/delegatecall/CallToENS.sol
guidanoli: It is 89 chars long, that's why `prettier` is breaking this line.
My suggestion is the following.

```suggestion
        bool success;
        bytes memory returndata;

        (success, returndata) = recipient.call{value: value}(payload);
```

test/foundry/dapp/Application.t.sol
```diff
@@ -64,6 +76,8 @@ contract ApplicationTest is ERC165Test {
     uint256 constant _initialSupply = 1000000000000000000000000000000000000;
     uint256 constant _tokenId = 88888888;
     uint256 constant _transferAmount = 42;
+    uint256[] _ERC1155IdArray = _generateUint256Array(7);
+    uint256[] _ERC1155SupplyArray = _generateUint256Array(7);
```
guidanoli, line 80: We're missing an array of transfer amounts so that we can actually test the balance after the transfer being non-zero. Also, I think we can come up with better names for these. Why not `_initialSupplies`, `_tokenIds`, and `_transferAmounts`?

Also, I'd be fine with having dynamic state variables only be initialized in `_initVariables`, just like `_interfaceIds`.  We could just loop over the number of token types, and push elements to these arrays one-by-one.

In my opinion, IDs can remain sequential (but then, `_generateUint256Array` wouldn't be necessary, as we'd have the index `i` readily available in the loop anyway, so we could remove that function), total supplies can be constant (and even equal to `_totalSupply`, a sensible default), and transfer amounts should be random and bounded by total supply.

test/foundry/dapp/Application.t.sol
```diff
@@ -359,6 +373,13 @@ contract ApplicationTest is ERC165Test {
         bytes memory output = _getOutput(name);
         OutputValidityProof memory proof = _getProof(name);
 
+        _testERC721(output, proof);
```
guidanoli, line 376: Why not `_testERC721Transfer`?

test/foundry/dapp/Application.t.sol
```diff
@@ -431,14 +452,22 @@ contract ApplicationTest is ERC165Test {
         bytes memory output = _getOutput(name);
         OutputValidityProof memory proof = _getProof(name);
 
+        _testERC20Fail(output, proof);
+        _testERC20Success(output, proof);
```
guidanoli, line 456: I don't think we need to separate these into two distinct functions.
We could have just `_testERC20Transfer`, which runs the failure case and then the success case.

test/foundry/dapp/Application.t.sol
```diff
@@ -359,6 +373,13 @@ contract ApplicationTest is ERC165Test {
         bytes memory output = _getOutput(name);
         OutputValidityProof memory proof = _getProof(name);
 
+        _testERC721(output, proof);
+    }
+
+    function _testERC721(
```
guidanoli, line 379: I understand you're aiming at minimal diffs for this PR for easier reviewing.
However, before merging, I think we should move the internal functions to the end of the contract.

test/foundry/dapp/Application.t.sol
```diff
@@ -497,6 +529,207 @@ contract ApplicationTest is ERC165Test {
         _appContract.executeOutput(output, proof);
     }
 
+    function testEtherTransferToENS() external _resolveENS {
```
guidanoli, line 532: This function could reuse a good chunk from the `testExecuteEtherTransferVoucher` function. Maybe move the code block from there to an internal function, and call it from here as well?

test/foundry/dapp/Application.t.sol
```diff
@@ -497,6 +529,207 @@ contract ApplicationTest is ERC165Test {
...
+
+    function testERC721TransferToENS() external _resolveENS {
+        string memory name = "ERC721ToENSDelegateCallVoucher";
+        bytes memory output = _getOutput(name);
+        OutputValidityProof memory proof = _getProof(name);
+
+        _testERC721(output, proof);
+    }
+
+    function testERC1155SingleTransferToENS() external _resolveENS {
```
guidanoli, line 608: I feel the lack of regular ERC-1155 transfer tests, without ENS.
Tip: define internal functions to avoid code duplication.

test/foundry/util/TestBase.sol
```diff
@@ -42,6 +42,16 @@ contract TestBase is Test {
         return array;
     }
 
+    function _generateUint256Array(
```
guidanoli, line 45: We don't need to add this function if we generate all three arrays in the same for-loop.

contracts/delegatecall/AssetTransferToENS.sol
```diff
@@ -0,0 +1,92 @@
...
+        bytes memory payload
+    ) external {
+        address recipient = resolveENS(node);
+
+        bool success;
+        bytes memory returndata;
+        (success, returndata) = recipient.call{value: value}(payload);
+        if (!success) {
+            returndata.raise();
+        }
```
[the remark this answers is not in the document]
ZzzzHui: We could create a library for `safecall`, although I think this code block is already easy to read on its own. Not sure if it improves readability by putting it into a library. Because when reading the codes, readers would need to jump to the library to see what the call is, and then jump back.

test/foundry/dapp/Application.t.sol
```diff
@@ -523,6 +756,9 @@ contract ApplicationTest is ERC165Test {
         _appOwner = _newAddr();
         _recipient = _newAddr();
         _tokenOwner = _newAddr();
+        _ens = ENS(_newAddr());
+        _resolver = AddrResolver(_newAddr());
+        _ensNode = keccak256(abi.encode(_recipient)); // random bytes32
```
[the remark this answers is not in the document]
ZzzzHui: `keccak256("user.eth")` looks good indeed

contracts/delegatecall/AssetTransferToENS.sol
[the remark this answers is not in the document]
ZzzzHui: If this is a library, then it helps to keep `node` as the first parameter, because then you can do `node.SomeTokenTransfer`. Although now it's not a library, it still acts as a library. And `node` is the "protagonist" of this contract. So I prefer keeping it first actually.

test/foundry/dapp/Application.t.sol
ZzzzHui: It has to be split into 2. Because in the test case `testERC20TransferToENS`, we need to call `_mockENS` between the 2. Because at the end of `_testERC20Fail`, it called `clearMockedCalls`

ZzzzHui: Although we could call `_mockENS` at the end of `_testERC20Fail` after `clearMockedCalls`, it's not the proper place for `_mockENS` to be in. So I don't mind splitting into 2.

test/foundry/dapp/Application.t.sol
```diff
@@ -497,6 +529,207 @@ contract ApplicationTest is ERC165Test {
...
+        assertTrue(
+            _wasOutputExecuted(proof),
+            "Output should be marked as executed"
+        );
+
+        _expectRevertOutputNotReexecutable(output);
+        _appContract.executeOutput(output, proof);
+    }
+
+    modifier _resolveENS() {
```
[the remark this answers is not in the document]
ZzzzHui: Only drawback is then we don't call `vm.clearMockedCalls` at the end of `_mockENS`, but I think it's ok. So I will adapt your suggestion

test/foundry/dapp/Application.t.sol
ZzzzHui: Yes. We should move internal functions to the end, unless it's easier to read at its current place. I think it might be easier to read here, but I don't mind moving it either

test/foundry/dapp/Application.t.sol
ZzzzHui: Agree, let's do that in another issue.

contracts/delegatecall/AssetTransferToENS.sol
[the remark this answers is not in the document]
ZzzzHui: changed my mind :)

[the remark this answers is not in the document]
guidanoli: There's some degree of strictness to which we should follow the DRY mantra ("don't repeat yourself"). I think in this case, it helps the user abstract the complexity of low-level operations such as `.call` and `.delegatecall` and handling of return data as error data, which is not very native Solidity-like. Wrapping that in a Solidity library helps abstract implementation details. Also, there's DRY, because this code is repeated here and in `Application`.

contracts/delegatecall/AssetTransferToENS.sol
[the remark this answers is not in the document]
guidanoli: The way I see it, the first parameter should be the contract you'll be calling, which, in this case, is still the token contract. We're just using `node` to resolve it to an Ethereum address, and be able to pass it to a function call to the token contract, therefore playing a secondary role.

test/foundry/dapp/Application.t.sol
guidanoli: It's just to keep things consistent, though I agree the function would be more distant to where it's being currently used.

contracts/delegatecall/AssetTransferToENS.sol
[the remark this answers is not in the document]
guidanoli: Having said that, in the specific case of `sendCallToENS`, I agree that `node` should be the first parameter, as it will resolve to an address which will be later called, just like a token contract in the other functions.

test/foundry/dapp/Application.t.sol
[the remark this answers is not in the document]
ZzzzHui: We don't need to. We used to not call `clearMockedCalls` at all, but started from some PR, we thought it's best practice to close them

contracts/delegatecall/AssetTransferToENS.sol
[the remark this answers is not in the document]
ZzzzHui: I get your perspective :) Our views differ in whether ENS node being a "primary" or "secondary" role. The reason why we have this contract is for various ENS specific cases. Think as if this is a library, and how `node` would be the first parameter if this was a library. I think the same logic still applies even this is not a library. 
Also, the order of the parameters is consistent with the order they are used within the function body too.
Let me know if you think it makes sense or doesn't.

test/foundry/dapp/Application.t.sol
ZzzzHui: Yeah we could do that too

test/foundry/dapp/Application.t.sol
```diff
@@ -523,15 +701,38 @@ contract ApplicationTest is ERC165Test {
...
+        _ens = ENS(_newAddr());
+        _resolver = AddrResolver(_newAddr());
         _interfaceIds.push(type(IApplication).interfaceId);
         _interfaceIds.push(type(IERC721Receiver).interfaceId);
         _interfaceIds.push(type(IERC1155Receiver).interfaceId);
+        for (uint256 i; i < 7; ++i) {
+            _tokenIds.push(i);
+            _initialSupplies.push(_initialSupply);
+            _transferAmounts.push(
+                bound(uint256(keccak256(abi.encode(i))), 0, _initialSupply)
```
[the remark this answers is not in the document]
ZzzzHui: I prefer keeping it this way, because:
1. the function doesn't get used elsewhere;
2. by keeping it in the current place, it's easier to see what's the range of the random amount, right?

test/foundry/dapp/Application.t.sol
```diff
@@ -523,15 +701,38 @@ contract ApplicationTest is ERC165Test {
...
+            );
+        }
+        if (_transferAmounts[0] == 0) {
+            // make the first transfer amount non-zero for easier tests
+            _transferAmounts[0] = 1;
+        }
```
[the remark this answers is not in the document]
ZzzzHui: I want to allow 0 `transferAmount` for other tokens. But keeping the first one non-zero helps with the revert case when the app balance is 0, see [here](https://github.com/cartesi/rollups-contracts/blob/43b13e38100a860f528723565d2f0bcfa4402ec3/test/foundry/dapp/Application.t.sol#L635). If we don't make the first transfer amount non-zero,  then it will not be the first token that reverts, it could be the 2nd or 3rd or 4th, which means we need a loop to know which token we expect to revert. Or if all transfer amounts are 0, then no revert will happen.
So I think it's easier to just make the first amount non-zero.

test/foundry/dapp/Application.t.sol
```diff
@@ -497,6 +537,144 @@ contract ApplicationTest is ERC165Test {
...
+            )
+        );
+        _appContract.executeOutput(output, proof);
+
+        vm.prank(_tokenOwner);
+        _erc1155BatchToken.safeBatchTransferFrom(
+            _tokenOwner,
+            address(_appContract),
+            _tokenIds,
+            _transferAmounts,
```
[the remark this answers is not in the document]
ZzzzHui: Yeah. We can do that too.

test/foundry/dapp/Application.t.sol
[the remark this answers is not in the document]
guidanoli: There is no reason to clear them if the test is going to end anyway after that.

contracts/delegatecall/AssetTransferToENS.sol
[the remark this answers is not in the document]
guidanoli: My point is that it shouldn't matter how you identify users, the action is an asset transfer. So, the only difference between an ERC-20 transfer to an Ethereum address and an ERC-20 transfer to an ENS name should be the recipient parameter, whose type changes from an `address` to a `bytes32`. So, from the application's point-of-view, they shouldn't care how this function is implemented, which variable is used first, etc. The application only cares about doing an asset transfer to an ENS name.

For example, let's say we want to transfer 1 CTSI to Alice.
If Alice is identified by her Ethereum address, `0xf39Fd...92266`, then the `DELEGATECALL` voucher would be...

```json
{
    "destination": "<SafeERC20Transfer>",
    "payload": {
        "functionSignature": "safeTransfer(address,address,uint256)",
        "arguments": [
            "<CTSI>",
            "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
            1000000000000000000
        ]
    },
}
```

Now, if Alice is identified by her ENS name, `alice.eth`, whose node hash is `0x78719...c3dec`, then I think it would be better if the argument order of the `DELEGATECALL` voucher should be similar.

```json
{
    "destination": "<AssetTransferToENS>",
    "payload": {
        "functionSignature": "sendERC20ToENS(address,bytes32,uint256)",
        "arguments": [
            "<CTSI>",
            "0x787192fc5378cc32aa956ddfdedbf26b24e8d78e40109add0eea2c1a012c3dec",
            1000000000000000000
        ]
    },
}
```

test/foundry/dapp/Application.t.sol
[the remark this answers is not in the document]
guidanoli: Given that transfer amounts are generated randomly, it is very unlikely that any one of them is zero. But if you want to be extra certain, you can just adjust the minimum transfer amount to be 1 rather than 0, [here](https://github.com/cartesi/rollups-contracts/blob/43b13e38100a860f528723565d2f0bcfa4402ec3/test/foundry/dapp/Application.t.sol#L712-L714).

guidanoli:
This needs a rebase.

contracts/delegatecall/AssetTransferToENS.sol
```diff
@@ -0,0 +1,86 @@
...
+    using LibAddress for address;
+    using SafeERC20 for IERC20;
+
+    ENS immutable _ens;
+
+    constructor(ENS ens) {
+        _ens = ens;
+    }
+
+    function sendCallToENS(
```
guidanoli, line 26: Maybe `sendMessageToENS` or `callENS`?

test/foundry/dapp/Application.t.sol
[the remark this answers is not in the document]
ZzzzHui: Yes. I thought the same.

test/foundry/dapp/Application.t.sol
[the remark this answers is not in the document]
ZzzzHui: Yes, we could do that too.

contracts/delegatecall/AssetTransferToENS.sol
ZzzzHui: since it can also be a simple ether transfer, `sendMessageToENS` probably doesn't fit. 
Between `sendCallToENS` and `callENS`, I think I prefer `sendCallToENS ` because all other functions in the contract have names like "sendxxToENS"

ZzzzHui:
rebased. Also moved internal functions to the end.

contracts/delegatecall/AssetTransferToENS.sol
guidanoli: I don't know. "Send call" just sounds weird to me, because you wouldn't normally say this in English. You either send something to someone, like a message, or you call them.

> since it can also be a simple ether transfer, sendMessageToENS probably doesn't fit.

I don't understand. Aren't Ether transfers simply message calls with a non-zero amount of Wei?

> because all other functions in the contract have names like "sendxxToENS"

That's one of the reasons why I wanted this function to be in another contract, because it's too general. It's not only suited for Ether transfers, but for arbitrary message calls. If we only care about Ether transfers, we could simply remove the `payload` argument and rename it as `sendEtherToENS`.

ZzzzHui: `sendEtherToENS` should cover most of the cases right? That was the main purpose.
And sending arbitrary calls was just a by-product.

guidanoli: Ok, let's rename it to `sendEtherToENS`. But should we keep the `payload` parameter? It would only be used to call a function of an ENS-identifiable contract, which is very specific, and unlikely, I presume.

ZzzzHui: Yes. Let's keep it. Users can decide what to do with it :)

test/foundry/dapp/Application.t.sol
```diff
@@ -426,68 +361,161 @@ contract ApplicationTest is ERC165Test {
...
         string memory name = "ERC20DelegateCallVoucher";
         bytes memory output = _getOutput(name);
         OutputValidityProof memory proof = _getProof(name);
 
-        assertLt(
-            _erc20Token.balanceOf(address(_appContract)),
-            _transferAmount,
-            "Application contract does not have enough ERC-20 tokens"
-        );
+        _testERC20Fail(output, proof);
```
guidanoli, line 369: Shouldn't this be `_testERC20Success`?

test/foundry/dapp/Application.t.sol
```diff
@@ -426,68 +361,161 @@ contract ApplicationTest is ERC165Test {
...
-        vm.expectRevert(
-            abi.encodeWithSelector(
-                IERC20Errors.ERC20InsufficientBalance.selector,
-                address(_appContract),
-                _erc20Token.balanceOf(address(_appContract)),
-                _transferAmount
-            )
-        );
-        _appContract.executeOutput(output, proof);
+        _testERC20Success(output, proof);
```
guidanoli, line 377: Shouldn't this be `_testERC20Fail`?

test/foundry/dapp/Application.t.sol
```diff
@@ -426,68 +361,161 @@ contract ApplicationTest is ERC165Test {
...
+    function testERC20TransferToENSFail() external {
+        string memory name = "ERC20ToENSDelegateCallVoucher";
+        bytes memory output = _getOutput(name);
+        OutputValidityProof memory proof = _getProof(name);
+
+        _testERC20Fail(output, proof);
+
+        _mockENS();
+
+        _testERC20Success(output, proof);
```
guidanoli, line 397: Did you mean to remove this? The successful case is already tested below.

test/foundry/dapp/Application.t.sol
```diff
@@ -426,68 +361,161 @@ contract ApplicationTest is ERC165Test {
...
+    }
+
+    function testERC20TransferToENSSuccess() external {
+        string memory name = "ERC20ToENSDelegateCallVoucher";
+        bytes memory output = _getOutput(name);
+        OutputValidityProof memory proof = _getProof(name);
+
+        _mockENS();
```
guidanoli, line 405: Did you mean to remove this? This function is already called at setup.

test/foundry/dapp/Application.t.sol
```diff
@@ -426,68 +361,161 @@ contract ApplicationTest is ERC165Test {
...
+                _erc1155BatchToken.balanceOf(
+                    address(_appContract),
+                    _tokenIds[i]
+                ),
+                _initialSupplies[i] - _transferAmounts[i],
+                "Application contract should have the transfer amount deducted"
+            );
+            assertEq(
+                _erc1155BatchToken.balanceOf(_recipient, _tokenIds[i]),
+                _transferAmounts[i],
```
guidanoli, line 514: I know that we _know_ that the application contract starts with `_initialSupplies[i]` and the recipient with zero tokens for each token type, but it would be better to construct `appBalances` and `recipientBalances` arrays before the transfer, and use them for the checks after the transfer.

guidanoli (commented):
Since we're allowing `sendEtherToENS` to receive a payload, I think it would be good to have a test where the ENS node actually resolves to the address of a contract, and the payload encodes a function call. We could even recycle parte of the "Ether mint" test case for that purpose.

test/foundry/dapp/Application.t.sol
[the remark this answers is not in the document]
guidanoli: Fair enough. :-)

test/foundry/dapp/Application.t.sol
ZzzzHui: Yes. Sorry for the silly mistake.

.changeset/olive-crabs-carry.md
```diff
@@ -0,0 +1,6 @@
+---
+"@cartesi/rollups": major
+---
+
+Added contract `AssetTransferToENS` that can be used as a destination for vouchers to transfer assets to ENS.
```
guidanoli, line 5: *for `DELEGATECALL` vouchers
*transfer assets to ENS-identified accounts

guidanoli (commented):
I've just noticed we're missing the deployment script of `SafeERC20Transfer` and `AssetTransferToENS`. They could be deployed by a `deploy/03_delegatecall.ts` script.

ZzzzHui:
> I've just noticed we're missing the deployment script of `SafeERC20Transfer` and `AssetTransferToENS`. They could be deployed by a `deploy/03_delegatecall.ts` script.

Yeah. But since `SafeERC20Transfer` is not relevant with this PR, the deploy script will be done in another PR. Ok? :)

guidanoli (approved):
Great job, LGTM!

ZzzzHui:
Needs @pedroargento 's approval to merge :)

pedroargento (approved):

PR #275, opened 2024-04-26
ENS portal

ZzzzHui:


contracts/portals/IENSPortal.sol
```diff
@@ -0,0 +1,26 @@
...
+    /// @notice ENS input can only be sent by its owner
+    error ENSOnlyOwner();
+
+    // Permissionless functions
+
+    /// @notice Send ENS input to InputBox
+    /// @param node The ENS node
+    /// @param appContract The application contract address
+    /// @param execLayerData Additional data to be interpreted by the execution layer
+    function sendENS(
```
guidanoli, line 21: Again, it's more than just the ENS node. You're telling the machine that the ENS node resolves to the Ethereum address of whoever sent that input.
Maybe we could rename it to something like `sendInputWithENSNode`?

contracts/delegatecall/AssetTransferToENS.sol
guidanoli, line 1: I see you're altering this file added by another PR.
Maybe add `LibENS` earlier in the PR chain, where it would be first used?

contracts/portals/ENSPortal.sol
```diff
@@ -0,0 +1,57 @@
...
+
+import {IENSPortal} from "./IENSPortal.sol";
+import {Portal} from "./Portal.sol";
+import {IInputBox} from "../inputs/IInputBox.sol";
+import {InputEncoding} from "../common/InputEncoding.sol";
+import {LibENS} from "../library/LibENS.sol";
+
+/// @title ENS Portal
+///
+/// @notice This contract allows anyone to send their ENS to the InputBox
```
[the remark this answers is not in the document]
ZzzzHui: If I understand you correctly, you are describing a scenario where someone will call this function regularly to send input blob. However, I think they should only call this function once to associate their address with the ENS node, and later on for the input blobs, call the regular `addInput` function. If there's any change to the ENS resolution, the next owner should call this function again. That would be more gas efficient, because we shouldn't be repeatedly doing ENS resolution when there's no change. Or am I missing something?

contracts/delegatecall/AssetTransferToENS.sol
ZzzzHui: I thought about that too. But in the previous PR, it wasn't necessary to make it a library. Since this PR also uses the same function, I move it into a library so we can re-use the code. Thus, it's ok to keep the change in this PR too, unless you think it's better to be in the previous PR.

contracts/portals/ENSPortal.sol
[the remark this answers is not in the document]
guidanoli: The idea is that the user will not be identified in the back-end through their Ethereum address, but rather through their ENS name. This way, the back-end never has an outdated view of the ENS registry, because resolution will occur at runtime when someone executes an output that deals with ENS node hashes.

contracts/delegatecall/AssetTransferToENS.sol
guidanoli: Alright.

contracts/portals/IENSPortal.sol
```diff
@@ -0,0 +1,26 @@
...
+pragma solidity ^0.8.8;
+
+import {IPortal} from "./IPortal.sol";
+
+/// @title ENS Portal interface
+interface IENSPortal is IPortal {
+    // Errors
+
+    /// @notice ENS input can only be sent by its owner
+    error ENSOnlyOwner();
```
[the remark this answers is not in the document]
ZzzzHui: "owner" and "resolution" should mean the same thing. But still, `AddressResolutionMismatch` does sound like a better name 👍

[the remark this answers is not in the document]
guidanoli: They are not necessarily the same. The owner of an ENS node is the entity that controls its settings and can make changes to its records, such as the associated Ethereum address or other settings. However, the Ethereum address to which a node resolves can be different from the owner's address.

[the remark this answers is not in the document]
ZzzzHui: I think you are talking about the owner of a resolver, rather than owner of an ENS name?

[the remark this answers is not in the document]
guidanoli: I know you are talking about the addresses that ENS names resolve to, but, according to [ENS terminology](https://docs.ens.domains/terminology), they are not necessarily the owners of the ENS names. It's different from, say, NFTs, where you actually own the tokens, and have full control over them.

Every ENS node has a [record](https://docs.ens.domains/terminology#record), which states the owner and the resolver. The [owner](https://docs.ens.domains/terminology#owner) can change the resolver, and transfer ownership of the node itself (e.g. `alice.eth`) and of sub-nodes (e.g. `pay.alice.eth`). The [resolver](https://docs.ens.domains/terminology#resolver) is the one actually responsible for resolving, for example, ENS nodes into Ethereum addresses.

Depending on the ownership regime of the resolver, the address that an ENS name resolves to can be changed by a single authority (see [`OwnedResolver`](https://github.com/ensdomains/ens-contracts/blob/staging/contracts/resolvers/OwnedResolver.sol)), or by a trusted ETH controller, or by a trusted reverse registrar, or by the owner of the node, or by an approved operator (see [`PublicResolver`](https://github.com/ensdomains/ens-contracts/blob/staging/contracts/resolvers/PublicResolver.sol)).

contracts/portals/IENSPortal.sol
```diff
@@ -0,0 +1,29 @@
...
+
+pragma solidity ^0.8.8;
+
+import {IPortal} from "./IPortal.sol";
+
+/// @title ENS Portal interface
+interface IENSPortal is IPortal {
+    // Errors
+
+    /// @notice ENS input can only ENS resolution matches the sender
```
guidanoli, line 12: How about...

"The provided ENS node does not resolve to the sender address."

contracts/portals/IENSPortal.sol
```diff
@@ -0,0 +1,29 @@
...
+    /// @param resolution The address that ENS resolves to
```
guidanoli, line 13: *the ENS node

contracts/portals/IENSPortal.sol
guidanoli, line 1: Maybe have a `getENS` function as well, just like we have `getInputBox`?

test/foundry/portals/ENSPortal.t.sol
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    function testSendInput(bytes calldata execLayerData) public {
+        bytes memory addInput = abi.encodeCall(
+            IInputBox.addInput,
+            (
+                _appContract,
+                InputEncoding.encodeENSInput(_ensNode, execLayerData)
+            )
+        );
```
guidanoli, line 95: Maybe have an `_encodePayload` and an `_encodeAddInput` function, like we do in other portal tests? I believe this might improve code readability.

guidanoli (approved):

PR #279, opened 2024-05-06
chore: deployment script addition

ZzzzHui:


deploy/03_delegatecall.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    const opts: DeployOptions = {
+        deterministicDeployment,
+        from: deployer,
+        log: true,
+    };
+
+    await deployments.deploy("SafeERC20Transfer", opts);
+
+    const ens = "0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e";
```
ZzzzHui, line 26: hey @guidanoli , curious about what you think is the best way to do this. This ENS registry address may not be the case for every network we want to deploy to.
(This also applies to the deployment of `ENSPortal`, which will be in the `01_input.ts` script)
Thanks

guidanoli: I don't know either. I think we need to research more about this topic.
Does such contract exist in every supported network? Or just Ethereum mainnet and testnet. If so, we might want to add a deployment script just for ENS-related contracts and work with this known address only for those networks. See https://docs.ens.domains/learn/deployments.

guidanoli: This article recommends to always refer to the ENS registry deployed to Ethereum mainnet.
https://makoto-inoue.medium.com/how-to-support-ens-for-multi-chain-dapps-b0a7ff043d77

guidanoli: I think it only makes sense to deploy ENS-related contracts to networks that we support and that the ENS DAO deploys their registries to. Therefore, `mainnet` and `sepolia`. We can skip a deployment script depending on the `hre` object by setting the `skip` attribute just like we set the `tags` attribute. See [`hardhat-deploy` documentation](https://github.com/wighawag/hardhat-deploy?tab=readme-ov-file#deploy-scripts).

ZzzzHui:
A new PR will be opened because of the recent changes to remove ENS related contracts.

PR #283, opened 2024-05-09
add convenience application factory for self-hosting

ZzzzHui:


contracts/dapp/ISelfHostedApplicationFactory.sol
```diff
@@ -0,0 +1,62 @@
...
+import {Authority} from "../consensus/authority/Authority.sol";
+import {IAuthorityFactory} from "../consensus/authority/IAuthorityFactory.sol";
+import {Application} from "./Application.sol";
+import {IApplicationFactory} from "./IApplicationFactory.sol";
+import {IInputBox} from "../inputs/IInputBox.sol";
+import {IPortal} from "../portals/IPortal.sol";
+
+/// @title Self-hosted Application Factory interface
+interface ISelfHostedApplicationFactory {
+    /// @notice Get the factory used to deploy `Authority` contract
```
guidanoli, line 15:
```suggestion
    /// @notice Get the factory used to deploy `Authority` contracts
```

contracts/dapp/SelfHostedApplicationFactory.sol
```diff
@@ -0,0 +1,97 @@
...
+        Application application = _applicationFactory.newApplication(
+            authority,
+            inputBox,
+            portals,
+            appOwner,
+            templateHash,
+            salt
+        );
+
+        return (application, authority);
```
guidanoli, line 72: You could make the return values named, just like you did in `calculateAddresses`, and just assign the newly-created contracts to them. With this, you won't need to declare them, won't need the `return` statement, and the first statement might even fit in one line?

guidanoli (commented):
Besides a small change, looks good!

contracts/dapp/SelfHostedApplicationFactory.sol
ZzzzHui: Hey Gui. I modified the `calculateAddresses` so there's no suffix `_` for return variables. I think suffix `_` for return variables make sense if there's prefix `_` for function parameters.  But we are using prefix `_` as internal state variables.

guidanoli: Oh, I didn't notice the `_` suffix. Yeah, we're no longer doing that. But I think we can still name return values, like `application` and `authority`.

ZzzzHui: True. Done. 👍

guidanoli (approved):
Great job, LGTM!

PR #295, opened 2024-07-26
Add last processed block number and epoch length

guidanoli:


contracts/consensus/authority/Authority.sol
```diff
@@ -17,14 +17,21 @@ contract Authority is AbstractConsensus, Ownable {
...
         address appContract,
+        uint256 lastProcessedBlockNumber,
         bytes32 claim
     ) external onlyOwner {
-        emit ClaimSubmission(msg.sender, appContract, claim);
-        _acceptClaim(appContract, claim);
+        emit ClaimSubmission(
+            msg.sender,
+            appContract,
+            lastProcessedBlockNumber,
```
ZzzzHui, line 32: I assume that we have the same philosophy here that validator node can do whatever it wants. So `lastProcessedBlockNumber` may not be incremental.
Means the `ClaimSubmission` and `ClaimAcceptance` events may be emitted with a smaller or same `lastProcessedBlockNumber` compared with previous events, with same/different claims.

contracts/consensus/IConsensus.sol
```diff
@@ -3,12 +3,20 @@
...
-/// @notice Validators may synchronize epoch finalization, but such mechanism is not specified by this interface.
+/// @notice Each application has its own stream of inputs.
+/// See the `IInputBox` interface for calldata-based on-chain data availability.
+/// @notice This stream of inputs is divided into epochs, with a given length.
+/// Each application defines its own epoch length.
+/// See the `getEpochLength` function in the `IApplication` interface.
+/// @notice We calculate the index of the epoch that an input belongs to by doing
+/// the integer division of the number of the base layer block in which the input
+/// was added by the epoch length:
+/// `epoch_index = base_layer_block_number / epoch_length`.
```
ZzzzHui, line 14: Do we start epoch from the base layer genesis, or the node may start with, for example, when the `InputBox`/dapp is deployed. If the node may start with a smaller number, then this formula here may cause confusion?

ZzzzHui (dismissed):

contracts/consensus/authority/Authority.sol
guidanoli: Yes, here the `lastProcessedBlockNumber` parameter is purely for debugging purposes, and not checked against anything on-chain.

contracts/consensus/IConsensus.sol
guidanoli: The node doesn't have to start reading events from genesis.
It can, for example, start reading events from the block in which the `InputBox` contract was deployed, as an optimization.
On the extreme, it can even start at the block in which the first `InputAdded` event was emitted for the given application.
Regardless of when the node starts reading events, epochs are defined from genesis for simplicity.
This does mean that the first epoch might have no inputs, but that's already the case if you pick any other base layer block to start with.
So, we chose to start with the genesis block to make this equation as simple as possible.

ZzzzHui: > The node doesn't have to start reading events from genesis.

Yeah I didn’t mean the node starts running from genesis. I meant the calculation of epochs starts from genesis.
As long as it’s consistent, it’s all good. And I agree starting from genesis makes calculations simpler 👍

marcelstanley (approved):

marcelstanley (approved):

ZzzzHui (approved):

PR #308, opened 2024-10-02
Remove v1 deployments

guidanoli:


tuler:
Is there any deployment left? Sepolia?

guidanoli:
From what I see, there is no deployment left.
The `export/abi` and `deployments` directories are gone.

tuler:
Oh, ok. Deroll's code needs at least 1 deployment to exist.

guidanoli:
Why does DeRoll need at least one deployment?

tuler:
> Why does DeRoll need at least one deployment?

Because it uses the contracts addresses, like as an example in the utility method `isERC20Deposit`

guidanoli:
Maybe you don't need at least one deployment.
You could have a mapping of trusted contracts indexed by chain ID.
Remember that the chain ID is now part of the input metadata schema.

```solidity
enum TrustedContractTypeID {
  Untrusted = 0,
  EtherPortal,
  ERC20Portal,
  ERC721Portal,
  ERC1155Portal,
}

// Could be implemented as a partial mapping instead of using
// 0 as a sentinel value for "untrusted"
mapping (uint256 chainId => mapping (address sender => TrustedContractTypeID)) trustedContractTypeID;

mapping (TrustedContractTypeID => AdvanceRequestHandler) trustedContractHandler;
```

Then, whenever it receives an advance request, it indexes this mapping with the chain ID and message sender from the metadata, and checks if it is trusted, and, if so, its type. From its type, it can forward it to a handler which can then decode it, etc.

So, without deployments info, every sender would be untrusted.

tuler:
> You could have a mapping of trusted contracts indexed by chain ID. Remember that the chain ID is now part of the input metadata schema.

Such mapping is being auto-generated by wagmi-cli, from deployment information in the npm package.
https://github.com/tuler/deroll/blob/main/packages/wallet/wagmi.config.ts

guidanoli:
So it could be an issue with the wagmi CLI or with the plugin.

tuler:
> So it could be an issue with the wagmi CLI or with the plugin.

There is no issue with the CLI or the plugin.
If there is no deployment, then there is no address to check on. And methods like `isERC20Deposit` would have to be removed.

guidanoli:
I reckon we're only removing v1 deployments to later add v2 deployments, right?
So, when we have v2 deployments, would Deroll work?

guidanoli:
I didn't have enough funds to deploy the contracts to Ethereum Sepolia, so I deployed them to OP Sepolia instead. Beware that this is just to have Deroll not break, and that contracts might still change before the definitive v2 release.

guidanoli:
Removed OP Sepolia deployment, because contracts changed in the meantime.

guidanoli:
Deroll will work because we have deployed the latest version of the contracts to Ethereum Sepolia.

tuler (approved):

PR #309, opened 2024-10-03
Make factories return interfaces

guidanoli:
Tagging everyone that partook in the thread.
Feel free to make any comments. :-)

### DIFF
--- a/test/foundry/util/OwnableTest.sol
+++ b/test/foundry/util/OwnableTest.sol
@@ -1,0 +1,114 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+pragma solidity ^0.8.22;
+
+import {IOwnable} from "contracts/access/IOwnable.sol";
+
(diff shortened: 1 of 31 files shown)


test/foundry/util/OwnableTest.sol
```diff
@@ -0,0 +1,114 @@
...
+
+        vm.startPrank(caller);
+        vm.expectRevert(
+            abi.encodeWithSelector(
+                Ownable.OwnableUnauthorizedAccount.selector,
+                caller
+            )
+        );
+        ownable.transferOwnership(randomOwner);
+        vm.stopPrank();
```
ZzzzHui, line 112: I feel like this is already done in the test `testUnauthorizedAccount`, but I guess we would never be too cautious right LOL

ZzzzHui (approved):

test/foundry/util/OwnableTest.sol
guidanoli: Yeah, we're being extra cautious here hah. :-)

PR #323, opened 2024-10-22
Upload Anvil deployment artifacts to GH Releases

### LINKED ISSUE #319: Add deployments file and post-deployment anvil state to the list of GitHub Release artifacts
## 📚 Context

In development environments, it is useful to have a local private network with all the contracts deployed in it. One way to achieve that is to run some development node (such as Anvil) and deploy the contracts through the deployment scripts.

One off-side to this approach is that the developer must have Node.js and `pnpm` installed in their local machine. Another option, if they are using Anvil as the development node, is to serialize the post-deployment state to a file.

This is an already supported feature of Anvil. Through the `--dump-state` option, you can serialize the state of the blockchain into a JSON file, which can be later recuperated by the `--load-state` command. If you want to load a state, do something, and then dump this new state into the same file, you can just use the `--state` command.

However, because Anvil is still under active development, and doesn't follow a strict semantic versioning approach, breaking changes can be introduced to the serialization format. Even recently, @tuler has improved Anvil by including mined transactions (see https://github.com/foundry-rs/foundry/pull/8411). Because of this, it might be wise to pin a specific version of Anvil and let it be clear on the name of state file that this version should be used. A later version might be compatible, but we won't hold any warranty.

## ✔️ Solution

On the CI, we can run `anvil --dump-state state.json` and `pnpm deploy:localhost` (after #317).
The resulting file can be renamed to incorporate the version of Foundry used, and uploaded as a GitHub Release artifact. Users can then download this artifact through the GitHub web UI or by downloading the file directly with `curl` or `wget`. On their side, they can then run `anvil --load-state state.json`.

Additionally, we should also add the `export/abi/localhost.json` (after #317) file to the list of GitHub Release artifacts so that users can more easily interact with the contracts, knowing their addresses and ABIs.

## 📈 Subtasks

- [ ] Edit CI to upload state file
- [ ] Release an RC to test the CI
- [ ] Check that state file indeed works

guidanoli:
Closes #319

### DIFF
--- a/src/tasks/deploy-anvil.ts
+++ b/src/tasks/deploy-anvil.ts
@@ -1,0 +1,60 @@
+import { spawn } from "child_process";
+import fs from "fs";
+import { task, types } from "hardhat/config";
+import { HardhatRuntimeEnvironment } from "hardhat/types";
+import path from "path";
+
+export interface DeployOptions {
+    dumpFile: string;
+    exportFile?: string;
+    silent: boolean;
+}
+
+const deploy = async (
+    taskArgs: DeployOptions,
+    hre: HardhatRuntimeEnvironment,
+) => {
+    const { dumpFile, silent } = taskArgs;
+    const exportFile =
+        taskArgs.exportFile ?? `export/abi/${hre.network.name}.json`;
+
+    // make sure directories exist
+    fs.mkdirSync(path.dirname(dumpFile), { recursive: true });
+    fs.mkdirSync(path.dirname(exportFile), { recursive: true });
+
+    // run anvil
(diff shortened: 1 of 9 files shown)


src/tasks/deploy-anvil.ts
guidanoli, line 1: Thanks to @tuler for letting me copy this from [`@cartesi/devnet`](https://github.com/cartesi/cli/tree/891d4298a3d8c3ba6b9a25b684c9427076efde23/packages/devnet).

pedroargento (approved):

PR #347, opened 2025-02-23
Add Cannonfile

guidanoli:
This PR just adds the Cannon CLI package as a dev dependency, and a Cannonfile.

### DIFF
--- a/cannonfile.toml
+++ b/cannonfile.toml
@@ -1,0 +1,78 @@
+name = "cartesi-rollups"
+version = "2.0.0"
+description = "Cartesi Rollups"
+
+[deploy.InputBox]
+artifact = "InputBox"
+create2 = true
+salt = "<%= zeroHash %>"
+ifExists = "continue"
+
+[deploy.EtherPortal]
+artifact = "EtherPortal"
+args = ["<%= contracts.InputBox.address %>"]
+create2 = true
+salt = "<%= zeroHash %>"
+ifExists = "continue"
+
+[deploy.ERC20Portal]
+artifact = "ERC20Portal"
+args = ["<%= contracts.InputBox.address %>"]
+create2 = true
+salt = "<%= zeroHash %>"
+ifExists = "continue"
+
+[deploy.ERC721Portal]
+artifact = "ERC721Portal"
+args = ["<%= contracts.InputBox.address %>"]
+create2 = true
+salt = "<%= zeroHash %>"
+ifExists = "continue"
+
+[deploy.ERC1155SinglePortal]
+artifact = "ERC1155SinglePortal"
+args = ["<%= contracts.InputBox.address %>"]
+create2 = true
+salt = "<%= zeroHash %>"
+ifExists = "continue"
+
+[deploy.ERC1155BatchPortal]
+artifact = "ERC1155BatchPortal"
+args = ["<%= contracts.InputBox.address %>"]
+create2 = true
+salt = "<%= zeroHash %>"
+ifExists = "continue"
+
+[deploy.AuthorityFactory]
+artifact = "AuthorityFactory"
+create2 = true
+salt = "<%= zeroHash %>"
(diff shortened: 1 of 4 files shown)


cannonfile.toml
```diff
@@ -0,0 +1,67 @@
+name = 'cartesi-rollups'
+description = 'Smart Contracts for Cartesi Rollups'
+version = '2.0.0'
+
+[deploy.InputBox]
+artifact = 'InputBox'
+create2 = true
+ifExists = 'continue'
```
tuler, line 8: Why do you need all these ifExists?

guidanoli: If, for some reason, `InputBox` is already deployed before running the script, the whole script doesn't fail, and instead just carries on. Because of the hash collision resistance properties of `keccak256` and how addresses are calculated with `CREATE2`, we can assume that if the pre-calculated address contains code, it is the same code that we're trying to deploy.

tuler: I think that has more potential to cause confusion than help users.
For example, some may think that they should run a load_state and then run cannon, which is not the expected process.

guidanoli: I don't see how setting `ifExists` may lead someone into thinking that this is the expected process. Also, to have a contract already deployed on a network before running the script is something perfectly possible to have in a real network, not just an Anvil private node. If anything, I believe this option only makes the Cannonfile more robust.

pedroargento:
We should on Node team opinion to see where they stand on adding this dependency. But lets wait for their release, they are really focused now.

cannonfile.toml
guidanoli: Another possibility is that we release a major version which changes some contracts, but not all of them. So, for example, we change the `ERC20Portal` but not the `InputBox` contract, we still want the script to carry on if there is already an `InputBox` contract deployed, right?

guidanoli:
> We should on Node team opinion to see where they stand on adding this dependency. But lets wait for their release, they are really focused now.

We could still distribute Anvil state dumps, along with deployment files (which contain contract addresses).

cannonfile.toml
tuler: > So, for example, we change the `ERC20Portal` but not the `InputBox` contract, we still want the script to carry on if there is already an `InputBox` contract deployed, right?

In that case it's another version, say 1.1.0, and in that you can define the InputBox as ifExists. But not in the 1.0.0 version.

I don't see why an error of running the 1.0.0 when it's already deployed is a problem.

cannonfile.toml
```diff
@@ -0,0 +1,67 @@
+name = 'cartesi-rollups'
+description = 'Smart Contracts for Cartesi Rollups'
+version = '2.0.0'
+
+[deploy.InputBox]
+artifact = 'InputBox'
+create2 = true
```
tuler, line 7: should we still use the safe-singleton-factory, or are there no more reasons?

I remember we switched to safe-singleton-factory was because it has much more networks supported, and back in time we had a more ambitious plan in terms of network support.

cannonfile.toml
guidanoli: > I don't see why an error of running the 1.0.0 when it's already deployed is a problem, or beneficial.

It's a problem, because the script is failing when, in fact, it shouldn't. It is fine to have a contract already deployed at that address, because we are using `CREATE2`.

cannonfile.toml
guidanoli: There are pros and cons to using Safe.

Pros:

- Can be deployed to networks that don't support legacy transactions

Cons:

- Not deployed by default on Anvil and Cannon local networks
- Depends on Safe team to sign transactions

Since we only support networks that support legacy transactions, I'd say there are no more pros to using Safe over Archnid.

tuler (dismissed):

tuler:
changeset?

tuler (approved):

PR #357, opened 2025-02-28
Remove unnecessary Node.js dependencies

guidanoli:
We can remove several Node.js dependencies:

- `hardhat` (replaced by `forge build`, for building the smart contracts)
- `hardhat-deploy` (replaced by `cannon build`, for deploying contracts)
- `solidity-coverage` (replaced by `forge coverage`, for generating coverage reports)
- `@openzeppelin/contracts` (now managed through `forge soldeer`)

With this, we're left with two Node.js dev dependencies:

- `@changesets/cli`, for managing releases
- `@usecannon/cli`, for managing deployments
- `prettier` (to be replaced by `forge fmt` in a future PR)


guidanoli:
TODO: export ABIs for Node.js package and GitHub Releases

pedroargento (approved):

foundry.toml
```diff
@@ -1,14 +1,26 @@
 [profile.default]
-src = 'contracts'
-out = 'out'
-libs = ['node_modules', 'lib']
-solc_version = '0.8.23'
+src = "contracts"
```
tuler, line 2: If this is a (almost) pure forge project, I'd keep the configuration file as small as possible, removing whatever can follow the default forge value.
Like src and out

foundry.toml
```diff
@@ -1,14 +1,26 @@
...
+out = "out"
+libs = []
+solc_version = "0.8.23"
 via_ir = true
 optimizer = true
+remappings = [
```
tuler, line 8: is this still necessary?

foundry.toml
guidanoli: This is planned, but I didn't want to pollute this PR.
Also, I wanted to ditch the `prettier` packages and use `forge fmt`.
However, their formatting differs every so slightly.
Even so, that would create some noise in the diff.
So, both moving `contracts` to `src` and removing `prettier` is adjourned to future PRs.

foundry.toml
guidanoli: Also temporary, just not to cause a huge diff in this PR.
I plan to use dependency versions, as it is intended by Soldeer, but for a future PR.

tuler (approved):

ZzzzHui (approved):

PR #358, opened 2025-03-02
fix: remove interface IDs from tests

guidanoli:
These interface ID arrays are reminiscent from when application and portal contracts implemented ERC-165.


ZzzzHui (approved):
Looks good!

PR #366, opened 2025-03-12
ci: update artifact upload workflow

guidanoli:
You can test the commands locally:

- `pnpm clean`
- `pnpm build:contracts`
- `pnpm build:devnet`

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,9 +2,12 @@
     "name": "@cartesi/rollups",
     "version": "1.2.0",
     "scripts": {
+        "build:contracts": "forge build contracts",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -2,9 +2,12 @@
     "name": "@cartesi/rollups",
     "version": "1.2.0",
     "scripts": {
+        "build:contracts": "forge build contracts",
+        "build:devnet": "cannon build --write-deployments deployments --anvil.dump-state state.json",
```
tuler, line 6: what about chain-id? does this build for 31337? Is the dump any different according to chain-id?

guidanoli: Building it for chain ID 31337 is more complicated.

PR #420, opened 2025-07-15
remove non deterministic deploy

ZzzzHui:


guidanoli (changes requested):
Looking good. I would add a changeset file for this **major** change:

```sh
pnpm changeset
```

The change can be something like the following:

```
Removed non-deterministic deployment functions from factories
```

ZzzzHui:
Thanks Gui. Changeset is added

guidanoli:
> Are you guys going to merge to main, or start a new alpha with changeset pre?

Good catch. Let's enter pre-release mode.

PR #432, opened 2025-07-23
add view function for number of executed outputs

ZzzzHui:


src/dapp/IApplication.sol
```diff
@@ -123,4 +123,7 @@ interface IApplication is IOwnable {
 
     /// @notice Get number of block in which contract was deployed
     function getDeploymentBlockNumber() external view returns (uint256);
+
+    /// @notice Get number of outputs that have been executed.
+    function numExecuted() external view returns (uint256);
```
guidanoli, line 128: How about `getNumberOfExecutedOutputs`?

test/dapp/Application.t.sol
```diff
@@ -569,6 +573,7 @@ contract ApplicationTest is Test, OwnableTest {
 
         _expectRevertOutputNotReexecutable(output);
         _appContract.executeOutput(output, proof);
+        assertEq(_appContract.numExecuted(), 1, "one output has been executed");
```
guidanoli, line 576: We should also test for more than one executed output.

guidanoli (approved):
Thanks, Zehui! You're always of great help. :blush:

ZzzzHui:
Thanks very much Gui 😊

PR #434, opened 2025-07-25
Add PRT Rollups Contracts

### LINKED ISSUE #411: Add PRT Rollups contracts
## 📚 Context

PRT is our first concrete fraud-proof algorithm integrated with Cartesi Rollups.
It enables anyone with modest funds and computing power to defend the truth.
To provide better support for PRT, we should have the PRT contracts here.
Currently, they are located in [`cartesi/dave`](https://github.com/cartesi/dave).
Later, when PRT matures, it may live in a separate repository.
It can even be Cartesi-agnostic, so that other rollups can use it for alternative VMs.

## ✔️ Solution

- Copy the PRT contracts.
- Copy the PRT unit tests.
- Add the PRT contracts to the root Cannonfile.

guidanoli:
Preserves commit history from [`cartesi/dave`](https://github.com/cartesi/dave).

Closes #411.

### DIFF
--- a/src/consensus/dave/DaveConsensus.sol
+++ b/src/consensus/dave/DaveConsensus.sol
@@ -180,6 +180,7 @@ contract DaveConsensus is IDataProvider {
         bytes32 realInputHash = _inputBox.getInputHash(_appContract, inputIndex);
         require(calculatedInputHash == realInputHash, InputHashMismatch(calculatedInputHash, realInputHash));
 
-        return input.getSmallestMerkleRootFromBytes();
+        uint256 log2SizeOfDrive = input.getMinLog2SizeOfDrive();
+        return input.getMerkleRootFromBytes(log2SizeOfDrive);
     }
 }
--- a/src/consensus/dave/DaveConsensus.sol
+++ b/src/consensus/dave/DaveConsensus.sol
@@ -3,59 +3,187 @@
 
 pragma solidity ^0.8.8;
 
+import {IInputBox} from "rollups-contracts/inputs/IInputBox.sol";
+import {Inputs} from "rollups-contracts/common/Inputs.sol";
+import {LibMerkle32} from "rollups-contracts/library/LibMerkle32.sol";
+
 import {IDataProvider} from "prt-contracts/IDataProvider.sol";
 import {ITournamentFactory} from "prt-contracts/ITournamentFactory.sol";
 import {ITournament} from "prt-contracts/ITournament.sol";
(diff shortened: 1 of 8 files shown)


src/consensus/dave/DaveConsensus.sol
```diff
@@ -0,0 +1,285 @@
...
+
+    /// @inheritdoc IOutputsMerkleRootValidator
+    function isOutputsMerkleRootValid(address appContract, bytes32 outputsMerkleRoot)
+        public
+        view
+        override
+        returns (bool)
+    {
+        require(
+            _appContract == appContract, ApplicationMismatch(_appContract, appContract)
```
ZzzzHui, line 245: I'm not sure if this check is necessary. Everything that's done in this consensus contract should all be for this one `_appContract`. So I was thinking this check should be done before doing anything (including this function and some other functions) in this consensus contract.
Maybe it's the use case of this function that requires this check to only be in this function?

src/consensus/dave/DaveConsensus.sol
```diff
@@ -0,0 +1,285 @@
...
+/// to know where epochs start and end, and which epochs have been
+/// settled already and which one is open for challenges still.
+/// Anyone can settle an epoch by calling `settle`.
+/// One can also check if it can be settled by calling `canSettle`.
+///
+/// @notice At any given time, there is always one sealed epoch.
+/// Prior to it, every epoch has been settled.
+/// After it, the next epoch is accumulating inputs. Once this epoch is settled,
+/// the accumlating epoch will be sealed, and a new
+/// accumulating epoch will be created.
```
ZzzzHui, line 43: For these 2 "will be", should we consider using "is"?

src/consensus/dave/DaveConsensus.sol
```diff
@@ -0,0 +1,285 @@
...
+    /// @param actual The actual epoch number in storage
+    error IncorrectEpochNumber(uint256 received, uint256 actual);
+
+    /// @notice Tournament is not finished yet
+    error TournamentNotFinishedYet();
+
+    /// @notice Hash of received input blob is different from stored on-chain
+    /// @param fromReceivedInput Hash of received input blob
+    /// @param fromInputBox Hash of input stored on the input box contract
+    error InputHashMismatch(bytes32 fromReceivedInput, bytes32 fromInputBox);
```
ZzzzHui, line 107: This error is not used in this contract. Also, should we say hash or merkle root, although merkle root is also some sort of hash as well haha

ZzzzHui (approved):

PR #441, opened 2025-09-19
Features/new view functions

pedroargento:


guidanoli (commented):
@pedroargento Looking good, just some minor cosmetic changes!

@vfusco Do you need a view method for the number of submitted claims as well, or just for the accepted ones?

test/consensus/authority/Authority.t.sol
```diff
@@ -118,6 +118,29 @@ contract AuthorityTest is Test, ERC165Test, OwnableTest {
         authority.submitClaim(appContract, lastProcessedBlockNumber, claim);
     }
 
+    function testSubmitMultipleValidClaims(uint8 numberOfClaims) public {
```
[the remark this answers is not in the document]
pedroargento: This change would increase the number of fuzzing tests because it would run for multiple randomly generated claims arrays, right? Since we’re not actually testing claims in this context, I’m not sure the additional runs add much value. What do you think?

src/consensus/AbstractConsensus.sol
```diff
@@ -16,7 +16,9 @@ abstract contract AbstractConsensus is IConsensus, ERC165 {
 
     /// @notice Indexes accepted claims by application contract address.
     mapping(address => mapping(bytes32 => bool)) private _validOutputsMerkleRoots;
-    uint256 _numberOfClaimsAccepted;
+    /// @notice Number of claims accepted by the consensus.
+    /// @dev Must be a A monotonically non-decreasing in time
```
guidanoli, line 20:
```suggestion
    /// @dev Must be monotonically non-decreasing in time
```

src/consensus/IConsensus.sol
```diff
@@ -80,5 +80,6 @@ interface IConsensus is IOutputsMerkleRootValidator {
     /// the integer division of the block number by the epoch length.
     function getEpochLength() external view returns (uint256);
 
-    function numberOfClaimsAccepted() external view returns (uint256);
+    /// @notice Get the claims accepted by the consensus.
```
guidanoli, line 83:
```suggestion
    /// @notice Get the number of claims accepted by the consensus.
```

src/dapp/Application.sol
```diff
@@ -51,6 +51,10 @@ contract Application is
     /// @dev See the `getDataAvailability` function.
     bytes internal _dataAvailability;
 
+    /// @notice The number of outputs executed by the application.
+    /// @dev See the `numberOfOutputsExecuted` function.
+    uint256 _numOfOutputsExecuted;
```
guidanoli, line 56: Should be renamed to keep consistency with v3.

```suggestion
    /// @dev See the `getNumberOfOutputsExecuted` function.
    uint256 _numOfExecutedOutputs;
```

test/consensus/authority/Authority.t.sol
```diff
@@ -118,6 +118,27 @@ contract AuthorityTest is Test, ERC165Test, OwnableTest {
...
 
+    function testSubmitMultipleValidClaims(bytes32[] calldata claims) public {
+        address owner = vm.addr(1);
+        address app = vm.addr(2);
+        uint256 epochLength = 5;
+
+        IAuthority authority = new Authority(owner, epochLength);
+
+        uint256 blockNum = 0;
+        vm.roll(blockNum);
```
ZzzzHui, line 129: these 2 lines seem not needed.

PR #462, opened 2025-11-13
local build without cannon

ZzzzHui:


foundry.toml
```diff
@@ -9,6 +9,7 @@ remappings = [
     "@openzeppelin-contracts-5.2.0/=dependencies/@openzeppelin-contracts-5.2.0/",
     "forge-std-1.9.6/=dependencies/forge-std-1.9.6/",
 ]
+fs_permissions = [{ access = "read-write", path = "out/"}]
```
guidanoli: I would change this to `path = "deployments/"` because `out` is the directory that Forge uses for its artifacts. Better not to avoid conflict.

scripts/Deploy.s.sol
```diff
@@ -0,0 +1,79 @@
...
+import "../src/portals/EtherPortal.sol";
+import "../src/portals/ERC20Portal.sol";
+import "../src/portals/ERC721Portal.sol";
+import "../src/portals/ERC1155SinglePortal.sol";
+import "../src/portals/ERC1155BatchPortal.sol";
+import "../src/consensus/authority/AuthorityFactory.sol";
+import "../src/consensus/quorum/QuorumFactory.sol";
+import "../src/dapp/ApplicationFactory.sol";
+import "../src/dapp/SelfHostedApplicationFactory.sol";
+import "../src/delegatecall/SafeERC20Transfer.sol";
```
guidanoli: It's good practice to import only what you need.
Please run `forge lint` to make sure there are no warnings.

```solidity
import {IInputBox} from "src/inputs/IInputBox.sol";
import {InputBox} from "src/inputs/InputBox.sol";
```

scripts/Deploy.s.sol
```diff
@@ -0,0 +1,79 @@
...
+
+contract Deploy is Script {
+    function run() external {
+        vm.startBroadcast();
+
+        bytes32 zeroSalt = bytes32(0);
```
guidanoli: This can be a script constant.

```solidity
bytes32 constant SALT = bytes32(0);
```

scripts/Deploy.s.sol
```diff
@@ -0,0 +1,79 @@
...
+        console2.log("EtherPortal: %s", address(etherPortal));
+        console2.log("ERC20Portal: %s", address(erc20Portal));
+        console2.log("ERC721Portal: %s", address(erc721Portal));
+        console2.log("ERC1155SinglePortal: %s", address(erc1155SinglePortal));
+        console2.log("ERC1155BatchPortal: %s", address(erc1155BatchPortal));
+        console2.log("AuthorityFactory: %s", address(authorityFactory));
+        console2.log("QuorumFactory: %s", address(quorumFactory));
+        console2.log("ApplicationFactory: %s", address(applicationFactory));
+        console2.log("SelfHostedApplicationFactory: %s", address(selfHostedApplicationFactory));
+        console2.log("SafeERC20Transfer: %s", address(safeERC20Transfer));
```
guidanoli: I think we can safely remove this logging, because addresses are being written to a JSON file anyways.

scripts/Deploy.s.sol
```diff
@@ -0,0 +1,79 @@
...
+        json = vm.serializeAddress("contracts", "ERC20Portal", address(erc20Portal));
+        json = vm.serializeAddress("contracts", "ERC721Portal", address(erc721Portal));
+        json = vm.serializeAddress("contracts", "ERC1155SinglePortal", address(erc1155SinglePortal));
+        json = vm.serializeAddress("contracts", "ERC1155BatchPortal", address(erc1155BatchPortal));
+        json = vm.serializeAddress("contracts", "AuthorityFactory", address(authorityFactory));
+        json = vm.serializeAddress("contracts", "QuorumFactory", address(quorumFactory));
+        json = vm.serializeAddress("contracts", "ApplicationFactory", address(applicationFactory));
+        json = vm.serializeAddress("contracts", "SelfHostedApplicationFactory", address(selfHostedApplicationFactory));
+        json = vm.serializeAddress("contracts", "SafeERC20Transfer", address(safeERC20Transfer));
+        vm.writeJson(json, "out/deployments.json");
```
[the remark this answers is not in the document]
guidanoli: I think we can remove the `out/`. Just write to `deployments.json` on the root of the project.

scripts/Deploy.s.sol
```diff
@@ -0,0 +1,79 @@
...
+        json = vm.serializeAddress("contracts", "EtherPortal", address(etherPortal));
+        json = vm.serializeAddress("contracts", "ERC20Portal", address(erc20Portal));
+        json = vm.serializeAddress("contracts", "ERC721Portal", address(erc721Portal));
+        json = vm.serializeAddress("contracts", "ERC1155SinglePortal", address(erc1155SinglePortal));
+        json = vm.serializeAddress("contracts", "ERC1155BatchPortal", address(erc1155BatchPortal));
+        json = vm.serializeAddress("contracts", "AuthorityFactory", address(authorityFactory));
+        json = vm.serializeAddress("contracts", "QuorumFactory", address(quorumFactory));
+        json = vm.serializeAddress("contracts", "ApplicationFactory", address(applicationFactory));
+        json = vm.serializeAddress("contracts", "SelfHostedApplicationFactory", address(selfHostedApplicationFactory));
+        json = vm.serializeAddress("contracts", "SafeERC20Transfer", address(safeERC20Transfer));
```
guidanoli: We can avoid some repetition by declaring a variable `obj` and assign it to something like `"contracts"`, as you're doing. Then, we can just pass `obj` to each function call. This also avoids passing the wrong string literal.

PR #466, opened 2025-11-15
Fix deployment Forge script

guidanoli:
## Problem

After #462 was merged, I decided to release an alpha version just so we could test the CI devnet build and it surely enough exposed a bug in the deployment Forge script: The `vm.getDeployment` cheatcode fails on simulation because it tries to access a broadcast file which does not exist. You see, this function basically attempts to retrieve the last deployment of a given contract to the current chain. So, on the first time you run the script, there is no previous deployment (during script simulation on a local fork of the network to which the contracts are being deployed). That is why the deployment Forge script, and, by extension, the devet build failed. This bug was not caught locally because my environment already contained a broadcast directory with transactions to the Anvil devnet (chain ID 31337). Had I cleaned my environment, or were the devnet build run on the PR CI, this bug would have been more likely caught at an earlier time.

## Solution

The solution is to not rely on broadcasts on first-time deployments. Since we are simulating both the deployment and the serialization of addresses in the same script entry-point, it would be simpler to simply store the addresses of deployments on variables that are later accessed for serialization. Straightforward, and less reliant on Forge cheatcodes. This is basically what @ZzzzHui had originally implemented, and was (mistakingly) changed by me to use `vm.getDeployment`.

### DIFF
--- a/.changeset/dull-geese-brush.md
+++ b/.changeset/dull-geese-brush.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/rollups": patch
+---
+
+Fix deployment Forge script
--- a/script/Deploy.s.sol
+++ b/script/Deploy.s.sol
@@ -18,58 +18,52 @@ import {ERC721Portal} from "src/portals/ERC721Portal.sol";
 import {EtherPortal} from "src/portals/EtherPortal.sol";
 
 contract DeployScript is Script {
+    /// @notice Salt used on deterministic deployments.
+    bytes32 constant SALT = bytes32(0);
+
     /// @notice Deploy the Cartesi Rollups contracts.
     /// @dev Serializes deployed contract addresses to `deployments.json`.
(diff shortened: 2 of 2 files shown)


ZzzzHui:
Hey Gui. What's the reason for going back to putting everything into basically one function?

guidanoli:
> Hey Gui. What's the reason for going back to putting everything into one function?

Hey, @ZzzzHui ! I've added an explanation to the PR body. Basically, we need to store deployments on variables to be later read during serialization. We use local variables rather than state variables because it is less verbose (declaration and assignment are perfomed by the same statement, removing the need to repeat the variable type) and avoids reading from uninitialized storage variables (which evaluate to zero). We could store the deployments on a struct in memory and make `_deployContracts` return it, and make `_serializeDeployments` receive it as a parameter, but that would create more boilerplate code unnecessarily.

Issue #478, opened 2026-01-16
IConsensus: add `getNumberOfSubmittedClaims()` view counter (for transition-based monitoring)


vfusco:
## Context

The node monitors two consensus events:
- `ClaimSubmitted`
- `ClaimAccepted`

We are migrating the node’s event monitoring to a transition-based approach (`FindTransitions`) that relies on monotonic on-chain counters exposed as `view` functions.

`IConsensus` already exposes a counter for accepted claims (`getNumberOfAcceptedClaims()`), but there is no equivalent counter for submitted claims.

## Request

Please add a `view` function to `IConsensus` (and implement it in the concrete consensus contracts) to return the total number of submitted claims.

This counter should be monotonic and represent the total number of `ClaimSubmitted` events emitted by the consensus.

## Why this matters

- Without a submitted-claims counter, the node cannot fully migrate `ClaimSubmitted` monitoring to the transition-based method.
- `ClaimSubmitted` is particularly important for node correctness on restart: the node needs to detect previously submitted claims to avoid resubmitting them after a crash/restart.

## References

- Existing accepted-claims counter in `IConsensus`: https://github.com/cartesi/rollups-contracts/blob/main/src/consensus/IConsensus.sol#L84

PR #479, opened 2026-01-16
Add monotonic counter for submitted claims

henriquemarlon:
Added getNumberOfSubmittedClaims() to IConsensus interface to expose a monotonic counter of submitted claims, as described in #478 

Behavior:
- Authority: 1 submission/increment per epoch, resubmission reverts;
- Quorum: 1 submission/increment per validator per epoch, same validator resubmission is silently ignored;

( I’ve also updated the tests ).

### DIFF
--- a/test/consensus/authority/Authority.t.sol
+++ b/test/consensus/authority/Authority.t.sol
@@ -133,6 +133,7 @@ contract AuthorityTest is Test, ERC165Test, OwnableTest {
             authority.submitClaim(app, blockNum - 1, claims[i]);
 
             assertEq(authority.getNumberOfAcceptedClaims(), i + 1);
+            assertEq(authority.getNumberOfSubmittedClaims(), i + 1);
         }
     }
 
@@ -172,6 +173,7 @@ contract AuthorityTest is Test, ERC165Test, OwnableTest {
         authority.submitClaim(appContract, lastProcessedBlockNumber, claim2);
 
         assertEq(authority.getNumberOfAcceptedClaims(), 1);
+        assertEq(authority.getNumberOfSubmittedClaims(), 1);
     }
 
     function testSubmitClaimNotEpochFinalBlock(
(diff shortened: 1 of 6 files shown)


test/consensus/authority/Authority.t.sol
```diff
@@ -133,6 +133,27 @@ contract AuthorityTest is Test, ERC165Test, OwnableTest {
...
+        assertEq(authority.getNumberOfSubmittedClaims(), 0);
+
+        for (uint256 i; i < claims.length; ++i) {
+            uint256 blockNum = (i + 1) * epochLength;
+            vm.roll(blockNum);
+
+            vm.prank(owner);
+            authority.submitClaim(app, blockNum - 1, claims[i]);
+
+            assertEq(authority.getNumberOfSubmittedClaims(), i + 1);
```
ZzzzHui: this block seems redundant? The behaviour is already tested in line 136, right?

test/consensus/quorum/Quorum.t.sol
```diff
@@ -458,6 +458,99 @@ contract QuorumTest is Test, ERC165Test {
...
+                vm.prank(quorum.validatorById(id));
+                quorum.submitClaim(claim);
+                ++totalSubmissions;
+                assertEq(quorum.getNumberOfSubmittedClaims(), totalSubmissions);
+            }
+
+            blockNum += epochLength;
+            vm.roll(blockNum);
+        }
+    }
```
ZzzzHui: since this is almost the same test as `testMultipleClaimsAcceptedCounter`, we could consider merging into it. We could then change the test name to something like `testMultipleClaimsCounters` etc

test/consensus/authority/Authority.t.sol
henriquemarlon: You’re right @ZzzzHui I made the adjustment. Let me know if it meets what you proposed.

ZzzzHui (approved):
LGTM!!

PR #484, opened 2026-02-19
Use GH CLI to upload assets to GH releases

guidanoli:
Context: https://github.com/softprops/action-gh-release/issues/704


guidanoli:
Once this PR is merged, we should release an alpha to test this change.

PR #496, opened 2026-03-25
Add version getter

guidanoli:
## Summary

This PR makes the following changes:

- Adds `IVersionGetter` interface, which defines a `version()` view function.
- Adds `RollupsContract` abstract contract, which implements `IVersionGetter`.
- Updates `RollupsContract.sol` on releases (similar to `cannonfile.toml`).
- Makes all main interfaces inherit from `IVersionGetter`.
- Makes all main contracts inherit from `RollupsContract`.
- Tests the version of all main contracts:
  - Ensures the version is a valid semantic version.
  - Ensures the version is in-sync with the `package.json` file.

Closes #490.

## Steps for reviewers

1. Update the `package.json` to not contain a `version` field. By running the
   script, you should see an error like `The package version is undefined`.

2. Similar to Step 2, but set the package version to a non-string value
   (`null`, a boolean, a number, or an object). The script should raise
   a slightly different error: `The package version is not a string`.

3. Update the `package.json` file and run `node ./scripts/update-version.mjs`.
   This should not only update `cannonfile.toml` (which is already expected)
   but also `src/common/RollupsContract.sol`. Check whether the update is sound.
   Repeat this for several versions until you're convinced.

4. Similar to Step 3, but with invalid versions. The script should raise an
   error like `The package version <VERSION> is not valid`.

5. Edit `src/common/RollupsContract.sol` to not contain one (or many) of
   the assignment statements. This will trigger an error on the script,
   since it cannot find such assignment in order to replace it.
   This is an expected behavior. We do not plan to make the script
   smart enough to inject an assignment if there is none yet, otherwise
   this would unnecessarily complicate the script.

6. Change the identation length of `src/common/RollupsContract.sol` and
   check that the script has honored it. (For example, try tabs or a
   different number of spaces).

7. Try running the script twice and check its idempotency property
   (It should print `Contracts version are already in-sync` on the
   second run and not alter the `src/common/RollupsContract.sol` file).

8. Check the coverage of the `test/util/VersionGetterTestUtils.sol` file.
   The test cases should cover all parsing machine states. You can see
   how to generate a code coverage report here:
   https://github.com/cartesi/rollups-contracts/pull/348

9. Check that all main interfaces inherit from `IVersionGetter`

10. Check that all main contracts inherit from `RollupsContract`

## Note to reviewers

We adopt the full semantic version format (`MAJOR.MINOR.PATCH[-PRE][+BUILD]`) instead of just `MAJOR.MINOR.PATCH[-PRE]` because we do not know yet whether `BUILD` would be useful in the future. For example, we could add the commit SHA or the version of the Solidity compiler as build metadata. Since this entrypoint will be used by the node to check whether a contract is compatible with a given SDK version, we should avoid brekaing changes in this entrypoint. Furthermore, if this function signature is adopted by other projects (such as `carteis/dave`), maybe they might want to add build metadata (for example, emulator version). So, that is why I think it can be beneficial if we used the full semantic version format.


GCdePaula (dismissed):
Looks good to me! Approved!

Minor typos only:

1. Typo: "reassambled" (x4)
`script/update-version.mjs:62,68,69` and `test/util/VersionGetterTestUtils.sol:39` -- should be "reassembled".

2. Typo: "occurence" (x6)
`script/update-version.mjs:80,92,104,116,128,157` -- should be "occurrence".

3. Grammar: "Contracts version are"
`script/update-version.mjs:175` -- should be "Contract versions are" or "Contracts version is".

4. == vs === for string comparison
`script/update-version.mjs:175,183` use == instead of ===. Both sides are strings so it works, but === is conventional JS style.

guidanoli:
@GCdePaula Thanks for spotting these typos! I'm force-pushing the patch.

GCdePaula (approved):
LGTM!!

PR #506, opened 2026-03-31
Convert error strings into custom errors

guidanoli:
Closes #492

### DIFF
--- a/src/withdrawal/IWithdrawalOutputBuilderErrors.sol
+++ b/src/withdrawal/IWithdrawalOutputBuilderErrors.sol
@@ -1,0 +1,14 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+pragma solidity ^0.8.8;
+
+interface IWithdrawalOutputBuilderErrors {
+    /// @notice This error is raised whenever a user provides an account
+    /// too short for the builder to decode. The error is accompanied by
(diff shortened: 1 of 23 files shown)


src/withdrawal/IWithdrawalOutputBuilderErrors.sol
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+pragma solidity ^0.8.8;
+
+interface IWithdrawalOutputBuilderErrors {
+    /// @notice This error is raised whenever a user provides an account
+    /// too short for the builder to decode. The error is accompanied by
+    /// the minimum expected account size suitable for on-chain decoding.
+    /// @param minAccountSize The minimum expected account size, in bytes.
+    error AccountTooShort(uint64 minAccountSize);
```
mpolitzer: Could we do something like this for a richer reporting?
```suggestion
    error AccountTooShort(uint64 attemptedAccountSize, uint64 minAccountSize);
```

mpolitzer (dismissed):

src/withdrawal/IWithdrawalOutputBuilderErrors.sol
guidanoli: The `account` would normally be part of the tx calldata, but we can definitely add this parameter, yes.

mpolitzer (approved):

PR #514, opened 2026-04-22
Claim staging

### LINKED ISSUE #513: Claim staging
## 📚 Specification

[Spec on HackMD](https://hackmd.io/@guidanoli/H1zvDHT3Zl)

## 📈 Subtasks

This to-do list chooses a test-driven development approach, mixing implementation and test tasks.

- [x] Add claim staging period parameter to consensus deployment entrypoints
- [x] Fix consensus deployment in unit tests
- [x] Add claim staging period getter to consensus interface and implementation
- [x] Check getter return value in unit tests
- [x] Add staged claim event definition to consensus interface and implementation
- [x] Add staged claim event counter to consensus interface and implementation
- [x] Check staged claim event counter returns zero initially
- [x] Add staged claim info getter to consensus interface and implementation
- [x] Check no claims are staged initially
- [x] Add claim staging internal function to abstract consensus implementation
- [x] Add claim acceptance function to consensus interface and implementation
- [x] Replace "accept claim" with "stage claim" in consensus implementations
- [x] Fix test cases that assume immediate claim acceptance
- [x] Check staged claim info, event and counter on claim submission (w/ and w/o staging)
- [x] Attempt to accept claim that was not staged
- [x] Attempt to accept claim that is staged but whose staging period is not over
- [x] Accept claim after staging period and check events/getters

guidanoli:
## Summary

This PR resolves #513 by implementing the [specified solution](https://hackmd.io/@guidanoli/H1zvDHT3Zl): When a claim is submitted by an Authority owner or by a Quorum majority, it is marked as staged; It can only be accepted after its staging period is elapsed.

## Note to reviewers

This PR attempts to make as little changes to the core contracts as possible, to make reviewing easier and reduce the likelihood of code defects. Tests were added to cover the additions and changes to the core contracts.

### DIFF
--- a/src/consensus/AbstractConsensus.sol
+++ b/src/consensus/AbstractConsensus.sol
@@ -27,8 +27,15 @@ abstract contract AbstractConsensus is
     /// @notice The epoch length
     uint256 immutable EPOCH_LENGTH;
 
-    /// @notice Indexes accepted claims by application contract address.
+    /// @notice The claim staging period
+    uint256 immutable CLAIM_STAGING_PERIOD;
+
+    /// @notice Indexes valid outputs Merkle roots by application contract address.
     mapping(address => mapping(bytes32 => bool)) private _validOutputsMerkleRoots;
+
+    /// @notice Indexes claim information by application contract address,
+    /// last-processed block number, and machine Merkle root.
+    mapping(address => mapping(uint256 => mapping(bytes32 => Claim))) private _claims;
 
     /// @notice Indexes number of the first unprocessed block
     /// by application contract address.
@@ -38,21 +45,25 @@ abstract contract AbstractConsensus is
     /// by application contract address.
     mapping(address => bytes32) private _lastFinalizedMachineMerkleRoots;
 
-    /// @notice Indexes number of claims accepted to the consensus
-    /// by application contract address.
+    /// @notice Indexes number of accepted claims by application contract address.
     /// @dev Must be monotonically non-decreasing in time
     mapping(address => uint256) private _numOfAcceptedClaims;
 
-    /// @notice Indexes number of claims submitted to the consensus
-    /// by application contract address.
+    /// @notice Indexes number of staged claims by application contract address.
+    /// @dev Must be monotonically non-decreasing in time
+    mapping(address => uint256) private _numOfStagedClaims;
+
+    /// @notice Indexes number of submitted claims by application contract address.
     /// @dev Must be monotonically non-decreasing in time
     mapping(address => uint256) private _numOfSubmittedClaims;
 
     /// @param epochLength The epoch length
+    /// @param claimStagingPeriod The claim staging period
     /// @dev Reverts if the epoch length is zero.
-    constructor(uint256 epochLength) {
+    constructor(uint256 epochLength, uint256 claimStagingPeriod) {
         require(epochLength > 0, IConsensusFactoryErrors.ZeroEpochLength());
         EPOCH_LENGTH = epochLength;
+        CLAIM_STAGING_PERIOD = claimStagingPeriod;
     }
 
     /// @inheritdoc IOutputsMerkleRootValidator
@@ -79,6 +90,10 @@ abstract contract AbstractConsensus is
         return EPOCH_LENGTH;
     }
 
+    function getClaimStagingPeriod() public view override returns (uint256) {
+        return CLAIM_STAGING_PERIOD;
+    }
+
     /// @inheritdoc IConsensus
     function getNumberOfAcceptedClaims(address appContract)
         external
@@ -87,6 +102,15 @@ abstract contract AbstractConsensus is
         returns (uint256)
     {
         return _numOfAcceptedClaims[appContract];
+    }
+
+    function getNumberOfStagedClaims(address appContract)
+        external
+        view
+        override
+        returns (uint256)
+    {
+        return _numOfStagedClaims[appContract];
     }
 
     /// @inheritdoc IConsensus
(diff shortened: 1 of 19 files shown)


stephenctw (commented):
Nice improvement overall. One small ask on tests: could we add 1-2 explicit “competing claim” cases to lock down semantics?

Same epoch + same machine root, different outputs root
Stage A, then submit B with same (app, lastProcessedBlockNumber, machineMerkleRoot) but different outputsMerkleRoot; assert expected behavior (no re-stage / no counter drift / no extra event).

Same epoch + different machine roots
Stage A and B for same (app, lastProcessedBlockNumber) with different machine roots; then accept one and assert what should happen to the other (acceptability, finalized root progression, and outputs validity).

Mostly for clarity + future-proofing: this makes exclusivity/canonicalization rules obvious from tests, instead of inferred from storage layout.

guidanoli:
> Same epoch + same machine root, different outputs root Stage A, then submit B with same (app, lastProcessedBlockNumber, machineMerkleRoot) but different outputsMerkleRoot; assert expected behavior (no re-stage / no counter drift / no extra event).

Hey @stephenctw! So, theoretically, the same machine Merkle root could be used to prove two different outputs Merkle roots, but we assume that crafting 59 Keccak-256 hash collisions is impossible. More precisely, we assume that if you prove two outputs Merkle roots OMR and OMR' for the same machine Merkle root, then OMR = OMR'.

**Keccak-256 collision-resistency hypothesis**\
$$\text{MMR}(\text{OMR}, p) = \text{MMR}(\text{OMR}', p') \implies \text{OMR} = \text{OMR}'$$

> Same epoch + different machine roots Stage A and B for same (app, lastProcessedBlockNumber) with different machine roots; then accept one and assert what should happen to the other (acceptability, finalized root progression, and outputs validity).

This test case is already being covered for both Authority (in [`AuthorityFactory.t.sol:493`](https://github.com/cartesi/rollups-contracts/blob/86551c0bcfa3f11297451313beb3525faf4fe175/test/consensus/authority/AuthorityFactory.t.sol#L493)) and Quorum (in [`QuorumFactory.t.sol:594`](https://github.com/cartesi/rollups-contracts/blob/86551c0bcfa3f11297451313beb3525faf4fe175/test/consensus/quorum/QuorumFactory.t.sol#L594)): If a validator attempts to submit a different claim for the same epoch, a `NotFirstClaim` error is raised. This ensures that at most one claim is staged per epoch.

stephenctw:
@guidanoli Thanks for the clarification! I must have overlooked the test case.

stephenctw (dismissed):
You can wait for @GCdePaula 's review/comment if you want. But this looks great to me.

guidanoli:
Thanks for the review, @stephenctw! Let us wait for @GCdePaula's review.

src/consensus/AbstractConsensus.sol
```diff
@@ -111,6 +135,67 @@ abstract contract AbstractConsensus is
...
+        require(
+            claim.status == ClaimStatus.STAGED,
+            ClaimNotStaged(
+                appContract, lastProcessedBlockNumber, machineMerkleRoot, claim.status
+            )
+        );
+        // We assume `block.number >= stagingBlockNumber` because when the claim is
+        // staged, we know `stagingBlockNumber` was assigned an evaluation of
+        // `block.number` in the current or in a previous block. Even if this assumption
+        // were wrong, the evaluation of the subtraction expression would raise a
```
GCdePaula: `a arithmetic underflow` -> `an arithmetic underflow`.

src/consensus/AbstractConsensus.sol
GCdePaula, line 1: Existing functions use `/// @inheritdoc IConsensus` (e.g. `getEpochLength`, `getNumberOfAcceptedClaims`), but the new functions don't (`getClaimStagingPeriod`, `getNumberOfStagedClaims`, `getClaim`, `acceptClaim`).

test/dapp/SelfHostedApplicationFactory.t.sol
```diff
@@ -109,6 +112,12 @@ contract SelfHostedApplicationFactoryTest is Test, VersionGetterTestUtils {
                 epochLength,
                 "authority.getEpochLength() != epochLength"
             );
+            assertGt(epochLength, 0, "getEpochLength() > 0");
```
GCdePaula, line 115: This assert is checking the `epochLength` fuzz parameter. Should this instead be `authority.getEpochLength()`?

GCdePaula (dismissed):
Great, solid PR!

I've left some very minor comments.

src/consensus/AbstractConsensus.sol
guidanoli: That is because I later found out that we don't actually need to use `@inheritdoc` if we aren't adding or changing any of the documentation, and that documentation is fully inherited by default. In fact, you can check this by running:

```sh
forge inspect AbstractConsensus userdoc | \
    jq -r '.methods["getClaim(address,uint256,bytes32)"].notice'
```

This should print out:

```
Get information about a claim.
```

Even though `AbstractConsensus.getClaim` doesn't define any documentation. It's all inherited from the `IConsensus` definition.

test/dapp/SelfHostedApplicationFactory.t.sol
guidanoli: In lines 110-114, it is asserted that `getEpochLength()` returns `epochLength`.

src/consensus/AbstractConsensus.sol
guidanoli: Patched in e2bab22, thanks!

guidanoli:
Hey, I've force-pushed a fix to the typo referenced in  https://github.com/cartesi/rollups-contracts/pull/514#discussion_r3144373372. You can just review the [diff](https://github.com/cartesi/rollups-contracts/compare/86551c0bcfa3f11297451313beb3525faf4fe175..e2bab226019775b6d2329ebd7e580107ce33024b).

GCdePaula (approved):

PR #515, opened 2026-04-28
Smaller account validity proof

### LINKED ISSUE #507: Smaller account validity proof
## 📚 Context

Since [3.0.0-alpha.2](https://github.com/cartesi/rollups-contracts/releases/tag/v3.0.0-alpha.2), we have added support to emergency withdrawals. However, the proof that users need to provide in order to validate their accounts and issue/execute withdrawal outputs is quite big. Considering single-leaf accounts, the proof length is 59. It doesn't need to be this big, since all proofs have the same suffix (given the same last-finalized machine state). Besides costing more gas, users would need to have access to the whole machine, and not just the accounts drive.

## ✔️ Solution

We can split the account validation process into two: First, anyone can validate the accounts drive Merkle root, given the last-finalized machine state (provided by the outputs Merkle root validator) and a proof. Second, users prove their accounts up to the accounts drive Merkle root, which is generally much smaller than the whole machine memory. For example, for a drive that supports $2^{17} = 131,072$ accounts, the proof size would drop from 59 to 17.

## 📈 Subtasks

- [x] Add a permisionless function called `proveAccountsDriveMerkleRoot`, which takes the accounts drive Merkle root and a proof as parameters, and validates them against the last-finalized machine state. This function can only be called after the app is foreclosed.
- [x] Add a view function called `getAccountsDriveMerkleRoot`, which returns whether the accounts drive Merkle root was proven and its value. Initially, it returns `(false, _)`, which effectively disallows withdrawals. Once `proveAccountsDriveMerkleRoot(x, proof)` is called, it should return `(true, x)`.
- [x] Make the `validateAccountMerkleRoot` function (which is used internally by `validateAccount` and `withdraw`) use the value returned by `getAccountsDriveMerkleRoot` to validate accounts Merkle roots, and require the proof to be the new (smaller) size.
- [x] Update documentation
- [x] Update unit tests to reflect the changes in the core contracts.

guidanoli:
## Summary

Currently, users need to prove their accounts through a full account-root-to-machine-root proof. However, we can split the account validation process into two steps: First, after the app is foreclosed, the accounts-drive Merkle root is proved against the last-finalized machine Merkle root. Second, accounts are validated based on the previously proved Merkle root of the accounts drive. The second proof is the one provided by the user, and should be much smaller than the whole account-root-to-machine-root proof. For example, for a hypothetical application that aims to support $2^{17}$ [USD accounts], the length of the account validity proof provided by the user would drop from 59 [^a] to 17 [^b], a 3.5x reduction. This reduces gas costs for users, but also enables slimmer withdrawal web UIs, as only the account recovery state file is required, and not the whole machine (assuming someone has altruistically proved the accounts-drive Merkle root).

This PR resolves #507.

## Note to reviewers

This PR attempts to make as few changes to the core contracts as possible, to make reviewing easier and reduce the likelihood of code defects. Tests were added to cover the additions and changes to the core contracts. As a suggestion, you can start reviewing the source contract changes, and then proceed to the test adaptations.

[USD accounts]: https://github.com/cartesi/rollups-contracts/blob/d7db66936779e7f9b53d12201a4dcccfae606c9f/src/library/LibUsdAccount.sol

[^a]: The machine Merkle tree height (59) minus log2 of the number of leaves per account (in this example, 0)
[^b]: The log2 of the maximum number of accounts (in this example, 17)

### DIFF
--- a/src/dapp/Application.sol
+++ b/src/dapp/Application.sol
@@ -44,6 +44,7 @@ contract Application is
     using LibAccountValidityProof for AccountValidityProof;
     using LibAddress for address;
     using LibBinaryMerkleTree for bytes;
+    using LibBinaryMerkleTree for bytes32[];
     using LibBytes for bytes;
     using LibOutputValidityProof for OutputValidityProof;
     using LibWithdrawalConfig for WithdrawalConfig;
@@ -94,6 +95,16 @@ contract Application is
     /// @notice Whether the application has been foreclosed by the guardian.
     /// @dev See the `isForeclosed` function.
     bool internal _isForeclosed;
+
+    /// @notice Whether the accounts drive Merkle root was proved.
+    /// @dev See the `getAccountsDriveMerkleRoot` and
+    /// `proveAccountsDriveMerkleRoot` functions.
+    bool internal _wasAccountsDriveMerkleRootProved;
+
+    /// @notice The accounts drive Merkle root.
+    /// @dev See the `getAccountsDriveMerkleRoot` and
+    /// `proveAccountsDriveMerkleRoot` functions.
+    bytes32 internal _accountsDriveMerkleRoot;
 
     /// @notice The number of outputs executed by the application.
     /// @dev See the `getNumberOfExecutedOutputs` function.
@@ -156,6 +167,44 @@ contract Application is
 
         ++_numOfExecutedOutputs;
         emit OutputExecuted(outputIndex, output);
+    }
+
+    function proveAccountsDriveMerkleRoot(
+        bytes32 accountsDriveMerkleRoot,
+        bytes32[] calldata proof
+    ) external override onlyForeclosed {
+        if (_wasAccountsDriveMerkleRootProved) {
+            revert AccountsDriveMerkleRootAlreadyProved();
+        }
+
+        if (
+            proof.length
+                != (CanonicalMachine.LOG2_MEMORY_SIZE - _getLog2AccountsDriveSize())
+        ) {
+            revert InvalidAccountsDriveMerkleRootProofSize();
+        }
+
+        // The Merkle root computation below should not raise an InvalidNodeIndex error
+        // because the LibWithdrawalConfig.isValid function run at the constructor
+        // guarantees that
+        // getAccountsDriveStartIndex() >> proof.length == 0.
+
+        bytes32 machineMerkleRoot = proof.merkleRootAfterReplacement(
+            getAccountsDriveStartIndex(), accountsDriveMerkleRoot, LibKeccak256.hashPair
+        );
+
+        // There is no risk of reentrancy attacks when retrieving the last-finalized
+        // machine Merkle root from the outputs Merkle root validator because it is done
+        // through a static call, which reverts on any state change.
+
+        bytes32 lastFinalizedMachineMerkleRoot = _getLastFinalizedMachineMerkleRoot();
+
+        if (machineMerkleRoot != lastFinalizedMachineMerkleRoot) {
+            revert InvalidMachineMerkleRoot(machineMerkleRoot);
+        }
+
+        _accountsDriveMerkleRoot = accountsDriveMerkleRoot;
+        _wasAccountsDriveMerkleRootProved = true;
     }
 
     function withdraw(bytes calldata account, AccountValidityProof calldata proof)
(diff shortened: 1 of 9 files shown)


src/dapp/Application.sol
```diff
@@ -260,22 +309,19 @@ contract Application is
...
-        bytes32 machineMerkleRoot = proof.computeMachineMerkleRoot(
-            accountMerkleRoot, getLog2MaxNumOfAccounts(), getAccountsDriveStartIndex()
-        );
+        bytes32 accountsDriveMerkleRoot =
+            proof.computeAccountsDriveMerkleRoot(accountMerkleRoot);
 
-        bytes32 lastFinalizedMachineMerkleRoot = _getLastFinalizedMachineMerkleRoot();
+        if (!_wasAccountsDriveMerkleRootProved) {
+            revert AccountsDriveMerkleRootNotProved();
+        }
```
GCdePaula, line 318: Hoisting this check before computing `accountsDriveMerkleRoot` feels more natural to me, in a "check before computation" kind of way

src/dapp/Application.sol
```diff
@@ -158,6 +169,44 @@ contract Application is
         emit OutputExecuted(outputIndex, output);
     }
 
+    function proveAccountsDriveMerkleRoot(
```
GCdePaula, line 172: Would it make sense to emit an event on this function?

test/dapp/Application.t.sol
GCdePaula: Could make sense to rename `testRevertsInvalidAccountIndex` -> `testRevertsInvalidNodeIndex`.

GCdePaula (commented):
Solid PR!

A comment on a pre-existing behavior that becomes more meaningful now.

The `migrateToOutputsMerkleRootValidator` is `onlyOwner` with no `notForeclosed` restriction. Combined with the new flow, this gives:

```
t1: guardian foreclose()                       // _isForeclosed = true; consensus state frozen
t2: nobody has called proveAccountsDriveMerkleRoot yet
t3: owner migrateToOutputsMerkleRootValidator(maliciousValidator)
t4: maliciousValidator.getLastFinalizedMachineMerkleRoot(app) returns attacker_chosen_value
t5: attacker calls proveAccountsDriveMerkleRoot(fakeRoot, craftedProof)
    where craftedProof + fakeRoot ⇒ attacker_chosen_value
t6: attacker constructs fake accounts whose Merkle root sits under fakeRoot, withdraws
```

Essential question: what is the threat model on the owner? Do we assume the owner is always trusted? Or is the guardian there to also protect against a malicious owner?

If this is indeed an issue, the mitigation is straightforward: gate `migrateToOutputsMerkleRootValidator` on foreclosure status.

guidanoli:
> Essential question: what is the threat model on the owner? Do we assume the owner is always trusted? Or is the guardian there to also protect against a malicious owner?

Hey @GCdePaula, you've raised a great point, and yes, this behavior is planned to be patched before the definitive v3 release. This feature is currently being tracked by issue #494. Since implementing this change won't affect the node directly, I decided to postpone this implementation to a later alpha release.

src/dapp/Application.sol
guidanoli: I mean, we could, but I'm not sure it's necessary, because if the node were to do a fast-sync on this event, it would use `getAccountsDriveMerkleRoot` either way, which already provides all the information necessary. Sometimes, an event is necessary because they contain information that is not necessary for the contract to function, but may be useful for off-chain usage. In this case, however, all information that is important related to this event is already obtainable through the view function. What do you think?

guidanoli: @vfusco Do you agree?

guidanoli:
@GCdePaula Addressed your first and third suggestions. The second one is pending a confirmation from the node team, whether we need an event for when the accounts-drive Merkle root is proved (or the view function is enough).

GCdePaula:
Ok! I'll tentatively approve.

On the second suggestion, all other methods emit an event, which is mainly why I flagged it.

I think regardless of the node needing it, might be good practice since it costs pretty much nothing.

I nevertheless think it's a useful event in this specific case; to know whether one needs to call the method for the first time, or if one can withdrawal.

GCdePaula (approved):

src/dapp/Application.sol
guidanoli: I've talked to Fusco and he said we move on without the event for now. If it proves itself necessary later down the line, we can add it before we release the definitive v3 release. :-)

PR #516, opened 2026-04-29
Fix Quorum not reverting with `NotFirstClaim` on claim resubmission

### LINKED ISSUE #495: Revert on Quorum same-claim resubmission
## 📚 Context

Currently, Quorum same-claim resubmission is allowed.
The transaction succeeds but emits no events.
This differs from Authority, in which `NotFirstClaim` is raised.
This divergence leads to more complicated Rollups node test code.

## ✔️ Solution

Make `Quorum.submitClaim` raise `NotFirstClaim` if a validator has already voted on the claim before.

## 📈 Subtasks

- [x] Update implementation
- [x] Update tests
- [x] Add changeset

guidanoli:
Currently, when a validators resubmits a claim to Quorum, the call is as a no-op. This differs from Authority, which reverts with a `NotFirstClaim` error. This PR makes Quorum behave like Authority in this aspect, so that the node tests don't need special treatment for Quorum.

Closes #495

### DIFF
--- a/.changeset/tricky-dryers-brake.md
+++ b/.changeset/tricky-dryers-brake.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/rollups": patch
+---
+
+Fix Quorum not reverting with `NotFirstClaim` on claim resubmission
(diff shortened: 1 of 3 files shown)


GCdePaula (approved):
LGTM!!

Issue #529, opened 2026-05-22
Type-check deterministic contract creation expressions in deployment script


guidanoli:
## 📚 Context

Currently, the deployment script deploys contracts through OpenZeppelin's `Create2` library:

```solidity
Create2.deploy(0, bytes32(0), initCode)
```

In the above function call, the only variable is `initCode`, which is a contatenation of two byte arrays:

```solidity
bytes memory initCode = abi.encodePacked(creationCode, encodedConstructorArgs);
```

Both byte arrays are provided as parameters:

- `creationCode`: Compile-time literal evaluation of `type(T).creationCode`
- `encodedConstructorArgs`: ABI-encoded constructor arguments

Notice how there is no way of type-checking constructor arguments.
The only way of detecting errors would be through deployment simulations.
In contrast, contract creation expressions are type-checked by the Solidity compiler.
Any faulty argument-passing would be caught at compilation time, not deployment/execution time.
For example:

```solidity
bytes32 salt;
InputBox inputBox = new InputBox{salt: salt}();
new EtherPortal{salt: salt}(inputBox); // ok
new EtherPortal{salt: salt}(); // error: missing argument!
new EtherPortal{salt: salt}(123); // error: invalid argument type!
new EtherPortal{salt: salt}(inputBox, inputBox); // error: extra argument!
```

This is a blind spot for the Solidity compiler, which can lead to unsound deployments.
Along with #528, this means we are not testing potentially unsound deployments.
This problem only affects deployment and (possibly) tests (if #528 is also addressed).

## ✔️ Solution

We need to have type-checked idempotent deterministic deployment contract creation in Solidity.
Unfortunately, neither Solidity nor Forge offer an idiomatic way of doing that for an arbitrary contract.
Even the latest Solidity version (0.8.35) does not offer generics, and Forge does not offer an idempotent CREATE2 back-end.
There is an issue on Foundry's repo about this: https://github.com/foundry-rs/foundry/issues/11485
Bottom line: We need to have, for each contract being deployed, a function like the following:
(Substitute `MyContract` and `uint256 arg` accordingly.)

```solidity
function deployMyContract(uint256 arg) external returns (MyContract myContract) {
    bytes32 salt;
    bytes memory creationCode = type(MyContract).creationCode;
    bytes memory encodedConstructorArgs = abi.encode(arg);
    bytes memory initCode = abi.encodePacked(creationCode, encodedConstructorArgs);
[...]

PR #533, opened 2026-06-08
Prepare for deposit refunds

### LINKED ISSUE #509: Fix documentation on `LibWithdrawalConfig`
## 📚 Context

https://github.com/cartesi/rollups-contracts/blob/67a3487272b50a764a44337ba4058902eba3866d/src/library/LibWithdrawalConfig.sol#L20

Thanks @lynoferraz for spotting this typo!

## ✔️ Solution

`type(uint8).max --> type(uint64).max`

### LINKED ISSUE #510: Bump machine-solidity-step from 0.13.0 to 0.14.0
## 📚 Context

The current v3 alpha is using machine-solidity-step 0.13.0.
However, machine-solidity-step 0.14.0 is available.
It is part of the Machine Emulator SDK 0.20.0 package.
We should bump this dependency to stay consistent with `cartesi/dave`.

## ✔️ Solution

Bump machine-solidity-step from 0.13.0 to 0.14.0.
See https://github.com/cartesi/dave/pull/241 for adaptations.

### LINKED ISSUE #528: Reuse deployment script code on tests
Currently, there are two pieces of code in which contracts are deployed:

- The **deployment script**: Serves to idempotently deploy the core contracts in any given chain and writes the deployment addresses to JSON files indexed by the target chain ID (e.g. `deployments/31337/InputBox.json` holds the deployment address of the `InputBox` contract for the Anvil development network, whose chain ID is 31337). The contracts are deployed through a `CREATE2` factory whose address is the same accross all chains that support legacy transactions (which is the case for all the supported networks).

- The **test contracts**: Serves to exhaustively test the core contracts. Currently, contracts in tests are deployed via the `CREATE` opcode (which would translate to contract-creating transactions when broadcasting to real networks).

This separation in deployment code leads to...

- Not testing the contracts that are going be deployed
- Code repetition

Yet, we don't want to just simply import the deployment scripts into the tests as this would make tests generate deployment artifacts.

### LINKED ISSUE #531: Discontinue Cannon support
## 📚 Context

There is an ongoing effort of migrating from Cannon to Forge deployment scripts.
Cannon has shown to be useful for some reasons, but unsuitable for others.
Forge scripts are way more flexible and reliable than Cannonfiles.
For simple deployments, Cannon is ellegant.
For more complex deployments, Cannon is a bottleneck.

## ✔️ Solution

Remove Cannonfile and any Cannon-related residue.

guidanoli:
This PR prepares the ground for the deposit refunds feature.
Here are the main changes:

- **Remove dependency from pnpm** (which includes Cannon and Changesets): We no longer depended on Cannon, and Changesets was used only internally. We will no longer publish the `@cartesi/rollups` package, and clients will need to download the Solidity compilation artifacts from GitHub releases. I would like to make releases immutable later to reduce even further the likelihood of supply-chain attacks. Currently, clients are expected to sumcheck artifacts from GitHub anyway, but immutable releases should add an extra layer of protection.

- **Ensure contract deployments are type-safe** (with code generation): Solidity doesn't currently support generics, so we resorted to encoding contract bytecode in a level too low for the Solidity compiler to type-check constructor arguments. We are now raising the abtraction level to `new` expressions with generated code so that Solidity can type-check contract deployments.

- **Reuse deployment code in tests** (follows DRY and increases test soundness): We are now testing the contracts exactly how they are being deployed (to Anvil devnet). If the tests are run in fork mode, you will test the contracts as if they were deployed to that network. Since contracts are the same accross all networks (except that devnet has more contracts for testing purposes), fork tests should not be necessary.

Before opening this PR, Claude and I have reviewed the changes twice.
Feel free to contact me if you have any questions!

Closes #509
Closes #510
Closes #528
Clsoes #529
Closes #531


.github/workflows/upload-rust-bindings.yml
```diff
@@ -49,4 +49,4 @@ jobs:
           fi
         env:
           DRY_RUN: ${{ !startsWith(github.ref, 'refs/tags/v') }}
-          CARGO_REGISTRY_TOKEN: ${{ startsWith(github.ref, 'refs/tags/v') && secrets.CARGO_REGISTRY_TOKEN || '' }}
\ No newline at end of file
+          CARGO_REGISTRY_TOKEN: ${{ startsWith(github.ref, 'refs/tags/v') && secrets.CARGO_REGISTRY_TOKEN || '' }}
```
brunomenezes, line 52: Just a comment. I kept staring at this change, not sure what really changed, maybe a tab or space.

guidanoli: The file was missing a newline character at the end.
Now the repo has no such files. You can check this by running:

```sh
git ls-files | while IFS= read -r f; do [ -f "$f" ] && [ -n "$(tail -c1 "$f")" ] && echo "$f"; done
```

Before this PR (on `next/3.0`), the command prints out:

```
.github/workflows/upload-contracts.yml
.github/workflows/upload-rust-bindings.yml
```

After this PR (on `feature/refactor`), it prints nothing.
You can even run this on other repos if you want.

brunomenezes (approved):
@guidanoli, it looks good to me. e.g. extraction of logic and reuse in the test files and etc.

PR #535, opened 2026-06-12
Refactor code and address issues #493, #530, #532, #534

### LINKED ISSUE #530: Use initial machine state for EW if no claim was accepted
## 📚 Context

For EW, the last-finalized machine Merkle root is used to prove the accounts drive Merkle root.
However, there is an edge case: What happens when no claim was accepted?
Currently, you wouldn't be able to prove the accounts drive Merkle root.
Why? Because the last-finalized machine Merkle root returned by the consensus would be zero.
(And you can't prove anything on a Merkle tree with a zeroed root.)

We could raise an error if no claim was accepted but...
- We don't know if app developers will deploy apps with non-empty EW state
- This adds friction in client software (node, CLIs, EW front-ends, etc)

So, a simpler alternative is to simply fallback to the template machine state.
On most cases, this will result in an empty EW state being proved, and no withdrawals.
But for those bizarre scenarios, we give support to EW.

## ✔️ Solution

Make `Application._getLastFinalizedMachineMerkleRoot()` return `getTemplateHash()` if `getLastFinalizedMachineMerkleRoot(address(this))` returns `bytes32(0)`. On the documentation of `getLastFinalizedMachineMerkleRoot(address)`, it is already specified that zero is supposed to be a sentinel value for "no claim". In the implementation of this function in `AbstractConsensus`, this is easily achieved from the fact that storage is a total mapping of EVM slot indices to EVM words with zero being the default EVM word value.

## 📈 Subtasks

- [x] Implement changes
- [x] Document edge case
- [x] Test edge case

### LINKED ISSUE #532: Deploy USD withdrawal output builder on devnet
## 📚 Context

For emergency withdrawal support, we created the concept of a withdrawal output builder.
We specified the interface and implemented a contract for accounts backed by USDC-like tokens.
In #519, we switched to deploying a factory for this contract instead of the contract itself.
That is because the contract takes the address of an ERC-20 token contract as constructor argument.
For devnet, such an address can be that of `TestFungibleToken`.
For other networks, we would have to elect a worthy ERC-20 token, like USDC.
However, this would violate the chain-independence property of Cartesi Rollups contract addresses.
However, we can still deploy the USD withdrawal output builder contract for devnet.
It would, in fact, be useful for tests on the Rollups Node codebase and for app developers.

## ✔️ Solution

Deploy `UsdWithdrawalOutputBuilder` (using `TestFungibleToken`) on devnet.

### LINKED ISSUE #534: Make foreclosure event single-emission
## 📚 Context

Currently, the `foreclose()` function can be called multiple times by the application guardian. This then leads to multiple `Foreclosure()` being emitted by the application contract, which is a weird behavior.

Moreover, the node stores the block in which the application was foreclosed. The foreclosure action is idempotent, but the event can be emitted multiple times. Making the event single-emission avoids this strange disparity.

This feature was requested by the node team.

## ✔️ Solution

Make `foreclose()` revert if `isForeclosed()` returns `true`.

## 📈 Subtasks

- [x] Update implementation
- [x] Update documentation
- [x] Test new behavior

guidanoli:
Continue the work of #533, this PR enriches the root Makefile, making it the reference for the project's name, version, Foundry version, as well as many other project-wide definitions. It also addresses several issues blocking the next alpha release:

- **Remove `IApplicationForeclosure` and `IApplicationWithdrawal` interfaces**: We realized this separation was not benefitial to the code organization. So, we merged them back into the `IApplication` interface.

- **Support withdrawals for applications with zero claims**: We used to assume that the machine initial state had an empty accounts drive, and therefore, it wouldn't be necessary to prove its Merkle root in order to validate accounts and execute withdrawal outputs. Now, if the application has zero claims, it uses the template hash when proving the accounts drive Merkle root, seamlessly covering this obscure edge case.

- **Make the `Foreclosure()` event single-emission**: This was a request by the node team. The `foreclose()` function now can only be called once by the guardian.

- **Bump Foundry from 1.4.3 to 1.5.1**: Benefits from improved test performance and fuzzer hability to finds bugs, new `cast erc20` commands, and single-line imports formatting. Clients can already test integration with the new devnet state dump by running `make devnet`.

- **Enrich Makefile and simplify workflow files**: The root Makefile now concentrates version definitions and recipes that make the workflows slimmer and more locally reproducible. The version file is now generated through a Forge script and from the version definition in the Makefile.

- **Deploy a test USD withdrawal output builder to devnet**: Essential for locally testing applications that leverage the initial emergency-withdrawal support for accounts backed by USDC-like tokens. This will simplify the cartesi CLI codebase (see [cartesi/cli#486](https://github.com/cartesi/cli/pull/486)).

-------------------------------

Closes #493
Closes #530
Closes #532
Closes #534


endersonmaia:
For consistency, I suggest using either the full chain name always or the shortname.

Or this, all short:

```
ARB_MAINNET
ARB_SEPOLIA
BASE_MAINNET
BASE_SEPOLIA
ETH_MAINNET
ETH_SEPOLIA
OP_MAINNET
OP_SEPOLIA
```

Or this, full name:

```
ARBITRUM_MAINNET
ARBITRUM_SEPOLIA
BASE_MAINNET
BASE_SEPOLIA
ETHEREUM_MAINNET
ETHEREUM_SEPOLIA
OPTIMISM_MAINNET
OPTIMISM_SEPOLIA
```

I prefer the first (short ones), where I use for the infra and also follow Alchemy's URL naming.

guidanoli:
Good idea. Let's use the short ones.

endersonmaia (approved):

PR #541, opened 2026-07-28
Machine yield check and other changes

### LINKED ISSUE #538: Machine yield check for Authority and Quorum
## 📚 Context

Currently, `AbstractConsensus.submitClaim` (which is inherited by both `Authority` and `Quorum`) receives an outputs Merkle root (which is assumed to be in the CMIO TX buffer) and its siblings in the machine. From these two parameters, it computes the machine Merkle root, which identifies the claim. However, in some cases, the CMIO TX buffer does not contain the outputs Merkle root. For example, the machine might have thrown an exception, in which case the app finds itself stuck in an unrecoverable state, and the application guardian would then act to foreclose the application, unlocking user funds through the recently included safety features: emergency withdrawal and deposit refunds.

In these cases, we would like for `submitClaim` to revert, meaning that validators cannot even submit a claim for a machine whose iflagsY register is not set and whose HTIF tohost register does not encode a "manual yield RX accepted" reason. One bad side effect of this design is that just by inspecting the blockchain state of the application, one cannot learn whether the validators are inactive or whether the post-epoch machine state is invalid and cannot be claimed. If a validator submit a claim for an invalid post-epoch machine state, the transaction would revert, so any events or state changes would be rolled back, and wouldn't be observable through `eth_call` or `eth_getLogs` JSON-RPC requests from the node. Instead, users would need to observe the application through a Rollups node instance to know whether the post-epoch state is invalid, so that the applciation guardian can foreclose the application.

## ✔️ Solution

Change the signature of the `IConsensus.submitClaim` function definition and adjust its implementation in the `AbstractConsensus` contract (which is already inherited by `Authority` and `Quorum`). Instead of receiving the outputs Merkle root and proof (which are used to compute the machine Merkle root), the function would now receive the machine Merkle root and Merkle proofs of three leaves: one with the iflagsY register, one with the HTIF tohost register, and the one (which is already being proved currently) with the first 32 bytes of the CMIO TX buffer (which in the case of a manually-yielded machine, is assumed to contain the outputs Merkle root).

```diff
 function submitClaim(
     address appContract,
     uint256 lastProcessedBlockNumber,
-    bytes32 outputsMerkleRoot,
-    bytes32[] calldata proof
+    bytes32 machineMerkleRoot,
+    MachineValidityProof calldata proof
 ) external;
```

Here is the structure of the machine validity proof. The data block field contains the 32 bytes of that machine memory range, which once hashed with Keccak-256, produces a leaf in the machine memory Merkle tree. Both iflagsY and HTIF tohost registers are unsigned 64-bit integers encoded in little-endian order and not aligned with 32 bytes, so the contracts need to do some decoding (in the soure code) and encoding (in the tests).

```solidity
struct LeafProof {
    bytes32 dataBlock;
    bytes32[] siblings;
}

struct MachineValidityProof {
    LeafProof iflagsYProof;
    LeafProof htifTohostProof;
    LeafProof txBufferProof;
}
```

## 📈 Subtasks

- [x] Implement and document the necessary changes in the source code
- [x] Adapt the existing test suite to the changes in the source code
- [x] Test invalid siblings array length
- [x] Test leaf proof producing different machine Merkle root
- [x] Test invalid iflagsY register value (zero)
- [x] Test invalid HTIF tohost register value (not manual yield RX accepted).

### LINKED ISSUE #539: Publish curated subset of compilation artifacts
## 📚 Context

On every release, we build all contracts under the `src` directory, and publish all compilation artifacts under the `out` directory in the form of a `cartesi-rollups-<version>-artifacts.tar.gz` bundle. These compilation artifacts contain all Rollups contracts ABIs, which are used to generate bindings in Go (for the Rollups node) and TypeScript (for the Rollups explorer and many other components). However, amongst these compilation artifacts are those from external dependencies (OpenZeppelin and the Cartesi Machine Solidity Step libraries) and internal contracts/interfaces. We can filter these compilation artifacts out when creating the bundle in order to reduce its size. The main benefits is that generating bindings for all contracts would take less time and yield much less code, and users might not even need to filter out contracts from the published artifacts (in the case of general-purpose bindings such as the Rollups TypeScript libraries for client/viem and react/wagmi users).

This feature was requested by both @brunomenezes and @tuler (as clients at the [rollups-ts](https://github.com/cartesi/rollups-ts) project).

## ✔️ Solution

Adjust the `tar` command in the compilation artifacts bundle Make target body to select a curated subset of all compilation artifacts, which encompasses only externally-facing interfaces and deployed contract implementations:

- ApplicationFactory
- AuthorityFactory
- Erc1155BatchPortal
- Erc1155SinglePortal
- Erc20Portal
- Erc721Portal
- EtherPortal
- IApplication
- IApplicationFactory
- IAuthority
- IAuthorityFactory
- IConsensus
- IErc1155BatchPortal
- IErc1155SinglePortal
- IErc20Portal
- IErc721Portal
- IEtherPortal
- IInputBox
- IOutputsMerkleRootValidator
- IQuorum
- IQuorumFactory
- IRefundOutputBuilder
- ISafeErc20Transfer
- ISelfHostedApplicationFactory
- IUsdWithdrawalOutputBuilder
- IUsdWithdrawalOutputBuilderFactory
- IWithdrawalOutputBuilder
- InputBox
- Inputs
- Outputs
- QuorumFactory
- RefundOutputBuilder
- SafeErc20Transfer
- SelfHostedApplicationFactory
- TestFungibleToken
- TestMultiToken
- TestNonFungibleToken
- UsdWithdrawalOutputBuilderFactory

With this filter, the bundle size went from 374KB to 189KB (half the original size).

## 📈 Subtasks

- [x] Filter compilation artifacts to bundle
- [x] Test new filter locally

### LINKED ISSUE #540: Encoding-agnostic account owner discoverability
## 📚 Context

Currently, the account encoding used by application is completely generic from the point of view of the `Application` contract. The only check it makes is that the account fits in the allocated slot in memory given by the withdrawal config constructor parameter. However, this leads to one problem: How to know the owner of an account encoded in a generic way?

## ✔️ Design and Solution

The first idea was to make the outputs withdrawal builder (which already needs to decode accounts in a specific encoding) return the owner address. However, this would result in $O(n)$ `eth_call` queries to the withdrawal output builder for $n$ accounts. A simpler approach would be for the withdrawal output builder to have an extra function that would return the offset after which the account owner occupies 20 bytes. This is much better. The node would only need to make a single `eth_call` in order to extract the owners of every account in the drive. This would already work, but there is an even simpler design: Have a known offset for every account encoding. This sounds limiting, but in actuality, for the USD-backed account encoding, this can be easily achieved by swapping some fields, and allows us to even expand the balance unsigned integer field from 64 to 96 bits, by consuming the existing padding field. From the options available, with a 32-bytes account size, and grounded in a hypothetical C++ implementation of the account struct with little-endian balances and aligned fields, we landed on a format in which the account owner occupies the last 20 bytes of the encoded account. And this is the standard that will be used for all accounts. If this standard no longer works, we can revisit it later in the future. Maybe have the offset function or some other design.

## 📈 Subtasks

- [x] Document the new requirement of the account owner occupying the last 20 bytes of the encoded account
- [x] Implement the changes in the USD account encoding/decoding implementation
- [x] Adapt the tests to use 96-bit unsigned integers for USD account balances

guidanoli:
## Summary

Addresses the last open issues in the [3.0.0-alpha.7 milestone](https://github.com/cartesi/rollups-contracts/milestone/9), and prepares the project for its release.

Closes #538, #539, and #540.

### Machine yield check for `Authority` and `Quorum` (#538)

Claims about post-epoch machine states can now only be submitted alongside proofs that the machine is manually yielded with an "RX accepted" reason.

- New `MachineValidityProof` struct bundling `LeafProof` fields for the `iflags_Y` register, the HTIF `tohost` register, and the first data block of the CMIO tx buffer (which contains the outputs Merkle root in the happy path).
- New `LibLeafProof` and `LibMachineValidityProof` libraries
- `IConsensus.submitClaim` (implemented by `Authority` and `Quorum`) takes `bytes32 machineMerkleRoot` and a `MachineValidityProof` in place of `bytes32 outputsMerkleRoot` and `bytes32[] proof`. The outputs Merkle root is now part of the machine validity proof rather than being supplied as a separate parameter.
- `_computeMachineMerkleRoot` and `_checkProofSize` internal functions drop out of `AbstractConsensus` in favour of `_validateMachine`, which returns the proven outputs Merkle root instead of receiving it. Diff on concrete contracts `Authority` and `Quorum` is minimal.
- `InvalidOutputsMerkleRootProofSize` is removed from `IConsensus`, which now inherits the new `MachineValidationErrors` interface (`InvalidSiblingsArrayLength`, `InvalidMachineMerkleProof`, `InvalidPostEpochMachineIflagsYRegister`, `InvalidPostEpochMachineHtifTohostRegister`).
- `CanonicalMachine` gains `DATA_BLOCK_MASK`, `IFLAGS_Y_ADDRESS` and `HTIF_TOHOST_ADDRESS`.

A machine that fails these checks may have reached an unrecoverable state, in which case the application should be foreclosed so users can recover funds through emergency withdrawals and deposit refunds.

### Encoding-agnostic account owner discoverability (#540)

Encoded accounts must now end with the 20-byte big-endian owner address, so the node can extract owners from the accounts drive and serve owner-to-account-index lookups without knowing the encoding.

- USD accounts are now exactly 32 bytes: `uint96` balance little-endian in the first 12 bytes (widened from `uint64`, consuming the former padding), owner in the last 20.
- `AccountTooShort(uint64, uint64)` becomes `InvalidAccountSize(uint256, uint64)` in `IWithdrawalOutputBuilderErrors`.
- The convention is documented on `IWithdrawalOutputBuilder`, which applies to any third-party builder, not just the USD one.

### Treat `ERC` as a word in `camelCase` and `PascalCase`

Mechanical rename motivated by Forge's Rust binding generation, which produced `ierc20_portal` instead of `i_erc20_portal`.

- Contracts and interfaces: `ERC20Portal` → `Erc20Portal`, `SafeERC20Transfer` → `SafeErc20Transfer`, `IERC1155SinglePortal` → `IErc1155SinglePortal`, and so on.
- Functions: `depositERC20Tokens` → `depositErc20Tokens`, `encodeERC20Deposit` → `encodeErc20Deposit`, etc.
- Errors: `ERC20TransferFailed` → `Erc20TransferFailed`.

For downstream clients this should be a find-and-replace of `ERC` with `Erc`.

### Publish a curated subset of compilation artifacts (#539)

`Makefile`-only change. The artifacts tarball now ships the same subset already used for Rust bindings generation, dropping dependency artifacts (OpenZeppelin, Machine Solidity Step) and internal contracts. Concrete deployed contracts stay in alongside their interfaces so `rollups-ts` can keep generating wagmi hooks.

### Supporting changes

- Bump `cartesi-machine-solidity-step` from 0.14.0 to the `0.15.0-test1` pre-release, which provides `EmulatorCompat.isYieldedManualWith` and the register addresses needed by #538. Another bump is expected once a definitive 0.15.0 tag lands.
- Merkle operations over a compressed node representation (test utilities), used to build machine Merkle roots with varying `iflags` and `tohost` values. `LibEmulator` is adapted to use it in place of the previous ad-hoc proof code.
- Move `ExternalLibBinaryMerkleTree` out of the test file into its own, renamed `LibBinaryKeccak256MerkleTree` to make the hash function explicit.
- `Makefile`: add `install-foundry`; add `check-foundry-version` as a prerequisite of `devnet`, so a state dump can no longer be produced with the wrong Foundry version; build RPC URLs automatically from `ALCHEMY_API_KEY`; use `addprefix` instead of `foreach`; drop an unused LCOV definition.

## Review notes

- `LibLeafProof` and `LibMachineValidityProof` are the security-sensitive parts of this PR — they decide whether a machine is deemed valid or not.
- The account encoding change is layout-only and silent at the ABI level: a 28-byte account built for alpha.6 will now revert with `InvalidAccountSize`, but a wrongly-ordered 32-byte one will decode into a garbage owner. Integration tests should be adapted to the new encoding.


guidanoli:
Tagging @GCdePaula @stephenctw for their knowledge of Solidity and Rollups contracts
Tagging @diegonehab @mpolitzer @edubart for their knowledge of the machine (register checks)
Tagging @brunomenezes @tuler for changes in packaging and ERC $\to$ Erc renaming
Tagging @endersonmaia for changes in contract deployment

tuler (approved):
Green for rename ERC casing. Note that it diverges from openzeppelin casing.

Green for packaging.

stephenctw (approved):
Solid on the account-encoding change (#540): uint96 LE + owner in the last 20 bytes looks correct, and the exact-32-byte check is clear.

One thing worth keeping visible for integrators: a 32-byte account with fields in the wrong order still decodes without reverting and yields a garbage owner. Length is enforced; layout is not. That matches the PR note, but third-party builders / apps need to get the byte order right — Application does not check owner-at-end.

edubart (approved):
The `LibMachineValidityProof.sol` handling of `IFLAGS_Y`/`HTIF_TO_HOST`/`TX_BUFFER` looks good for me. The `LibUsdAccount.sol` handling of `uint96` integers also looks good for me.

src/library/LibMachineValidityProof.sol
```diff
@@ -0,0 +1,44 @@
...
+    using LibLeafProof for LeafProof;
+
+    /// @notice Validate a machine and prove its outputs Merkle root.
+    /// @param v The machine validity proof
+    /// @param machineMerkleRoot The machine Merkle root
+    /// @return outputsMerkleRoot The proven outputs Merkle root
+    /// @dev May raise `InvalidSiblingsArrayLength`, `InvalidMachineMerkleProof`,
+    /// `InvalidPostEpochMachineIflagsYRegister`, or
+    /// `InvalidPostEpochMachineHtifTohostRegister`.
+    function validate(MachineValidityProof calldata v, bytes32 machineMerkleRoot)
```
GCdePaula, line 24: We verify  `iflags_Y != 0` and the HTIF device, command, and reason, then returns the entire first TX block as the outputs root, right? It could make sense to, in addition, assert that the low 32-bit HTIF data field (which contains the CMIO payload length) is 32.

GCdePaula (approved):
LGTM! Left one minor comment.

src/library/LibMachineValidityProof.sol
guidanoli: Yes, but machine-solidity-step currently doesn't export the `HTIF_DATA_{MASK,SHIFT}` constants.

PR #544, opened 2026-08-04
Add test USDC token

### LINKED ISSUE #543: Add a test USDC token to devnet
## 📚 Context

Applications often use USDC and need a test token on devnet to develop. The devnet already has a TestFungibleToken but the number of decimals is different, 18 instead of 6.

The workaround is still use the TestFungibleToken, but not use the onchain decimals, hardcoding in the application a "fake" 6 value, but wallets like Metamask use what is onchain, so it's confusing for the user testing the application.

## ✔️ Solution

Deploy a token similar to TestFungibleToken but with 6 decimals. The symbol is less relevant, but I don't see a problem of using USDC, even for a test.

guidanoli:
Changes:

- Add `TestUsdc` contract, an ERC-20 token contract with 6 decimal places, similar to USDC
- Deploy `TestUsdc` to devnet
- Deploy `TestUsdWithdrawalOutputBuilder` with `TestUsdc` (instead of `TestFungibleToken`)
- Adjust tests (use `TestUsdc` instead of `TestFungibleToken` for EW-related stuff)

Closes #543

Note for @vfusco: This might require some changes the rollups-node EW integration tests, because we'll use this test USDC rather than the test fungible token in the test USD withdrawal output builder contract deployed to devnet.

### DIFF
--- a/.github/workflows/upload-artifacts.yml
+++ b/.github/workflows/upload-artifacts.yml
@@ -21,6 +21,8 @@ jobs:
 
       - name: Build release artifacts
         run: make -j release-artifacts
+        env:
+          ALCHEMY_API_KEY: ${{ secrets.ALCHEMY_API_KEY }}
 
       - name: Upload assets to release on GitHub
         if: startsWith(github.ref, 'refs/tags/v')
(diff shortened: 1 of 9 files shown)


tuler (dismissed):

guidanoli:
Sorry, @tuler. Forgot to include `TestUsdc` in the list of contracts whose compilation artifacts we publish on releases, since we agreed to include all deployed contracts in this curated subset.

tuler (approved):

PR #549, opened 2026-08-06
ci: bump ci actions

endersonmaia:
`npx actions-up`

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -9,7 +9,7 @@ runs:
       shell: bash
       run: echo "foundry-version=$(make print-foundry-version)" >> $GITHUB_OUTPUT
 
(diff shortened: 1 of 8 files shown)


guidanoli (commented):
Apparently, `upload-pages-artifact` v4.0.0 excludes dotfiles from the artifact by default. The `forge doc -b` command generates a `.nojekyll` file, whose contents are:

```
This file makes sure that Github Pages doesn't process mdBook's output.
```

However I'm not sure whether this file is necessary or not, since the docs CI worfklow only runs on `main`.

guidanoli (approved):
Commit SHA-1 hashes and tags match.
Action bumps are backwards-compatible.
(The `.jekyll` file is ignored but should not break the docs.)

PR #550, opened 2026-08-06
Add `verify-*` Make targets

guidanoli:
Changes:

- Add a `verify-<chain>` target for every supported livenet, which reads the deployment artifacts of that chain and verifies its contracts through `forge verify-contract` (on Sourcify by default, or on the chain's Etherscan-based explorer if `ETHERSCAN_API_KEY` is set)
- Add aggregate `verify-testnets`, `verify-mainnets` and `verify-livenets` targets
- Add `VERIFY_OPTS`, counterpart of `DEPLOY_OPTS`
- Add a `DeploymentSummary` Forge script, which writes one deployment per line to `deployments/summary.txt` (chain ID, contract name and address), filtered per chain with `awk`
- Generate the `deploy-*` and `verify-*` recipes with GNU Make macros
- Add `docs/verification.md`, and update `docs/deployment.md`

Tested against all Sepolia-based tests.
You can review it too by running the `{deploy,verify}-testnets` targets on your side.
Since all contracts were deployed and verified, these targets will succeed silently.

If you want to test fresh deployment and verification, make dummy changes to source code (e.g. change the project version on the Makefile and run `make codegen`; this will affect all contracts that are deployed to livenets).

Closes #548

PR #551, opened 2026-08-12
Prepare for 3.0.0-alpha.8

### LINKED ISSUE #542: Fix Soldeer pkg release CI job
## 🙂 Expected behavior

Expected `.github/workflows/upload-contracts.yml` to publish the contracts to Soldeer package `cartesi-rollups-contracts` on `v*` tag pushes.

## 🫠 Actual behavior

The CI job did not publish the package to Soldeer. Instead, it merely dry-ran the publishing process, as it is done on other triggering events (such as PRs).

## 🧪 Evidence

See the [`upload-contracts.yml`@`v3.0.0-alpha.7`](https://github.com/cartesi/rollups-contracts/actions/runs/30906004110/job/91981101664) GitHub Actions run:

```
Run make publish-soldeer-package
  make publish-soldeer-package
  shell: /usr/bin/bash -e {0}
  env:
    DRY_RUN: false
    SOLDEER_API_TOKEN: ***
┌  🦌 Soldeer Push 🦌
│
├  Running in dry-run mode, a zip file will be created for inspection
│  
●  Zip file created at /home/runner/work/rollups-contracts/rollups-contracts/rollups-contracts.zip
│  
└  Done!
```

Notice how the `DRY_RUN` environment variable evaluated to `false`.
On the root Makefile, the dry-run mode is activated if `DRY_RUN` is non-empty.
Thus, the `publish-soldeer-package` Make target ran in dry-run mode.

## 🌎 Environment

GitHub Actions runner.
Cartesi Rollups Contracts 3.0.0-alpha.7.

## ✔️ Possible solutions

- Change the Make target to accept falsy values such as `false` and `0`.
- Change the GitHub workflow file to evaluate `DRY_RUN` to empty on `v*` tag pushes.

### LINKED ISSUE #547: Bump machine-solidity-step to 0.15.0
Machine Solidity Step 0.15.0 is [out](https://github.com/cartesi/machine-solidity-step/releases/tag/v0.15.0)!
We're currently using the `v0.15.0-test1` tag, but it's good to use `v0.15.0` for the definitive v3 release.

### LINKED ISSUE #548: Post-deployment verification
## 📚 Context

Verification ensures good user experience on Etherscan for auditors/devs/end-users.
Deployers can verify contracts on deployment with the `--verify` deployment option.
However, sometimes verification fails post-deployment and there is no way to attempt verification later.
Upstream, https://github.com/foundry-rs/foundry/issues/10593 discusses possible solutions to this issue.
In the meantime, we need a workaround that enables post-deployment verification.

## ✔️ Solution

Deployment artifacts are currently stored in the `deployments` directory.
We can use such artifacts to verify smart contracts on Etherscan after deploying them.
Through the `forge verify-contract` command, we can attempt such verification.
To the rollups-contract developer, these actions would be Makefile targets.
Similar to the `deploy-*` family, we would have the following `verify-*` targets:

- `verify-<chain>` (where `chain` $\ne$ `devnet`)
- `verify-testnets`
- `verify-mainnets`
- `verify-livenets` (livenets $=$ testnets $\cup$ mainnets)

Ideally, these targets would enable verification for both Etherscan and Sourcify.
Also, to avoid extra dependencies, it would be ideal to use Forge scripts instead of `jq` for JSON processing.
At the end of the day, there should be a straightforward way to verify contracts on releases.
This verification process should be well documented under `docs`.

## 📈 Subtasks

- [x] Add `verify-*` Makefile targets
- [x] Test new feature on Ethereum Sepolia
- [x] Document verification process
- [ ] Validate solution with @endersonmaia

guidanoli:
## Changes

- Add post-deployment verification through `verify-*` targets
- Write deployment addresses in plaintext, and deprecate the JSON artifacts
- Bump `cartesi-machine-solidity-step` to the definitive `v0.15.0` tag
- Fix `publish-soldeer-package` for the release workflow
- Bump the version and update the `CHANGELOG`

This supersedes #550, which can be closed. The implementation changed: instead of a `DeploymentSummary` script producing a single `deployments/summary.txt` filtered with `awk`, each deployment is written to its own `deployments/<chain-id>/<contract-name>.txt`. This lets each `verify-<chain>-<contract>` target take its artifact as a prerequisite, so Make itself reports a missing deployment (`No rule to make target ...`) instead of the target silently verifying nothing.

The Makefile is now `.NOTPARALLEL`, and `-j` was dropped from the release-artifacts workflow. Parallel Forge invocations race on the compilation cache and on the solc downloads performed by svm-rs, which was the real reason the `deploy-*` targets depended on `build`; that dependency is gone now.

Closes #542
Closes #547
Closes #548

## Review notes

- Deployment and verification was tested against Ethereum Sepolia. To exercise it yourself, run `make deploy-testnets` followed by `make verify-testnets`. It will dry-run and skip seamlessly in the case of Ethereum Sepolia. You may test it on other chains or change the version and run `make codegen` so that the addresses change.
- The release bundles are now reproducible byte-for-byte, so a rebuild of the same revision should produce identical tarballs.
- JSON deployment artifacts are deprecated but still written, so nothing breaks today. Please migrate `rollups-node` (@vfusco), `rollups-ts` (@tuler) and `rollups-explorer` (@brunomenezes) to the `.txt` files, so we can drop the JSON ones in a later release.


Makefile
```diff
@@ -1,157 +1,199 @@
...
+.PHONY: verify-testnets
+
+# ------------------------------------------------------------------------------
+# Auxiliary definitions
+# ------------------------------------------------------------------------------
 
 NOOP  =
 SPACE = $(NOOP) $(NOOP)
 
+TO_UPPER = $(subst a,A,$(subst b,B,$(subst c,C,$(subst d,D,$(subst e,E,\
```
tuler, line 29: Jesus 😅

tuler (approved):

Makefile
guidanoli: Yeah, I know... Maybe a bit overkill, but I honestly don't mind.

guidanoli: The alternative would be to invoke shell commands, which IMO is worse.

guidanoli: We could also list screaming-case conversions for each chain.
[screaming-snake-case-refactor.patch](https://github.com/user-attachments/files/30988205/screaming-snake-case-refactor.patch)

brunomenezes (approved):

Makefile
tuler: That won't change the fact I hate Makefile and think they are awful. 😅
My initial comment was like "Jesus, Makefiles are really horrible".
You can leave it as is. Humans won't probably touch that anyway.
But you have to agree [just](https://just.systems) is a step forward to Makefiles, and it has an uppercase function.

guidanoli: > Humans won't probably touch that anyway.

What do you mean? Are you saying that perhaps I'm not human? :joy:

> That won't change the fact I hate Makefile and think they are awful. 😅
> My initial comment was like "Jesus, Makefiles are really horrible".
> But you have to agree [just](https://just.systems/) is a step forward to Makefiles, and it has an uppercase function.

I like `just`, but we already depend on the GNU toolchain, which features Make.
Now, as a matter of personal preference, I don't _love_ Makefiles, but I also don't _hate_ them.

tuler: > Are you saying that perhaps I'm not human? 😂

I cannot infer that anymore.

> I don't _love_ Makefiles, but I also don't _hate_ them.

We don't have that in common.

PR #555, opened 2026-08-18
Prepare for 3.0.0-alpha.9

guidanoli:
Bumps `PRE_RELEASE` to `alpha.9` in the Makefile and regenerates `src/common/Version.sol` via `make codegen`. Also makes several Makefile improvements and collects the changes made since alpha.8.


endersonmaia (approved):
LGTM

